PR #8, opened 2025-02-03
Add a Flink adapter and simple translator

ayirr7:
https://github.com/getsentry/streaming-planning/issues/12

A really initial attempt at having a runtime-agnostic translation step. Extend StreamAdapter to add new runtimes. 
 
Still need to:
- See how generic the translation step is, perhaps by adding a `BytewaxAdapter` (https://github.com/getsentry/streaming-planning/issues/16)
- Decide if we like this approach of mapping our primitives/APIs to primitives/APIs in existing runtimes
- Need to add more primitives (like `Map`) to flesh out FlinkAdapter and see how usable all of the current abstractions are
- Config is littered across various files right now, we should make clear which configs are used and where they're set


py/sentry_streams/example_config.py
```diff
@@ -1,16 +1,16 @@
-from sentry_streams.sinks import Pipeline, RawKafkaSink, RawKafkaSource
+from sentry_streams.pipeline import Pipeline, RawKafkaSink, RawKafkaSource
 
 # pipeline: special name
 pipeline = Pipeline()
 
-source = RawKafkaSource(name="myinput", ctx=pipeline, logical_topic="logical-events")
+source = RawKafkaSource(
+    name="myinput", ctx=pipeline, logical_topic="logical-events", step_type="source"
```
ayirr7, line 7: The idea behind adding `step_type` was to give us a way to map our primitives API (e.g. `RawKafkaSource`) to a canonical primitive name (e.g.`Source`). We'd do this for other primitives too if we like this way of mapping them. 

Open to other ways of mapping as well. I could also use the dataclass names to figure out which runtime primitives they correspond to.

py/tests/test_pipeline.py
```diff
@@ -0,0 +1,65 @@
...
+from sentry_streams.adapters.stream_adapter import StreamAdapter
+from sentry_streams.pipeline import Pipeline, RuntimeTranslator
+from sentry_streams.runner import iterate_edges
+
+
+# Essentially identical to runner for now
+# Checks basic input and output stream types
+# of a simple Flink program that uses
+# Pipeline abstractions
+def test_pipeline() -> None:
```
ayirr7, line 15: A first test, kind of an integration test. It feels to me that this is the most useful test for the current state of the repo. Most of the functionality in the repo so far is to support a simple e2e pipeline like the one this test has.

py/sentry_streams/adapters/flink_adapter.py
fpacifici, line 1: Let's make a package per platform. If we start having more than one I doubt the entirety of the platform adapter will fit in one module.

py/sentry_streams/example_config.py
fpacifici: Isn't a `RawKafkaSource` always a source as well as a `RawKafkaSink` is always a sink? 
If this is correct and we need to classify whether a step is a sourceor a sink, that should be a structural part of the class rather than an argument to the constructor. 

A few ways to do that are:
- provide a static property for each of the class that returns the type, or, much better, return what is needed by the client to chaneg its behavior depending whether this is a source or a sink/
- Extend a class that contains everything that makes a class a source/sink

py/sentry_streams/pipeline.py
```diff
@@ -0,0 +1,109 @@
...
+        translated_fn = getattr(self.adapter, next_step)
+
+        return translated_fn(step_config, stream)
+
+
+class PipelineGraph:
+    def __init__(self) -> None:
+        self.steps: dict[str, Step] = {}
+        self.incoming_edges: dict[str, list[str]] = defaultdict(list)
+        self.outgoing_edges: dict[str, list[str]] = defaultdict(list)
```
fpacifici, line 32: nit: `MutableMapping` rather than `dict`

py/sentry_streams/pipeline.py
```diff
@@ -0,0 +1,109 @@
...
+        self.sources: list[Source] = []
+
+    def register(self, step: Step) -> None:
+        assert step.name not in self.steps
+        self.steps[step.name] = step
+
+    def register_edge(self, _from: Step, _to: Step) -> None:
```
fpacifici, line 39: nit: why were `from` and `to` called `_from` and `_to` ?

py/sentry_streams/runner.py
```diff
@@ -32,56 +73,23 @@ def main() -> None:
...
+    for source in p_graph.sources:
         print(f"Apply source: {source.name}")
-        env_source = source.apply_source(env, environment_config)
+        env_source = source.apply_source()
         step_streams[source.name] = env_source
-        iterate_edges(step_streams)
+        iterate_edges(step_streams, p_graph)
 
     # submit for execution
     env.execute()
```
fpacifici, line 95: Shouldn't the transaltor return an `env` object as that is Flink specific ?

It owuld be alright if we kept the runtime translation for a separate PR

py/sentry_streams/pipeline.py
```diff
@@ -0,0 +1,109 @@
...
+@dataclass
+class RawKafkaSink(WithInput):
+    logical_topic: str
+    step_type: str
+
+    def apply_edge(self, stream: Any) -> Any:
+        step_config = {"topic": self.logical_topic}
+        sink_output = self.ctx.translator.translate_with_input(self, step_config, stream)
+
+        return sink_output
```
fpacifici, line 97: Could you please clarify the reason for this design:
1. the translator is a property of the pipeline
2. each step has an apply method that calls the translator passing itself.
3. The step does not navigate to the next step in the chain.

As each step class is fairly light regarding its logic. It basically only carry properties. Wouldn't it be simpler to make the adapter read the step object and build the translated data structure ? Then the translator could navigate the graph, and, for each step found it would identify the right adapter method, pass the step to the adapter method and wire up the results.

The pattern in this PR (add a generic apply method on the step and pass the translator as an object) reminds a lot the visitor pattern, but the key there is that the step itself would navigate to its children and invoke the visitor there. 
Here the graph is navigated by external code, so the advantage of the visitor does not seem to be there.

On the other hand the visitor pattern would work if we called `accept` passing the visitor to the source and the source navigated through the graph delegating the followign step to the outbound edges.

fpacifici (changes requested):
Some high level comment. 

I think, at this point, I would keep this a little simpler as we do not have a clear picture of the right abstraction and in order to  make it easier to iterate.

I think one of these options may be better:
- (preferred as it is easier to iterate) leave the iteration on the runner for now. and make the runner call the transaltor method accordingly passing the step to the translator. The translator wires up the result of the translate method to the stream.
- (more complex, i think ti should be a followign iteration) use a proper visitor to navigate the graph and build the Flink structure. This would mean hiding the Flink environment inside the visitor as well.

py/sentry_streams/runner.py
ayirr7: It seems that with Flink, all you need to do is register the source with the `env`. (We do that in the `FlinkAdapter`) 

Then, subsequently traversing the graph and translating each step just yields a new DataStream after each step. You don't need to pass around`env` as you traverse the graph. 

All you need is a call to `env.execute()` to actually submit the stream to be run.

py/sentry_streams/pipeline.py
ayirr7: Not sure but can change

ayirr7: I think it was written that way because `from` is reserved keyword

fpacifici: oh right

py/sentry_streams/runner.py
```diff
@@ -1,9 +1,54 @@
...
+    Pipeline,
+    RuntimeTranslator,
+    WithInput,
+)
+
+
+# Traverses over edges in a PipelineGraph, building the
+# stream incrementally by applying steps and transformations
+# It currently has the structure to deal with, but has no
+# real support for, fan-in and fan-out streams
```
fpacifici, line 18: please make it a docstring

py/sentry_streams/runner.py
```diff
@@ -1,9 +1,54 @@
...
+
+                # 1:1 between input and output stream
+                else:
+                    next_step: WithInput = cast(WithInput, p_graph.steps[output_step_name])
+                    print(f"Apply step: {next_step.name}")
+                    next_step_stream = translator.translate_step(next_step, input_stream)
+                    step_streams[next_step.name] = next_step_stream
+                    output_stream = next_step_stream
+
+    return output_stream
```
fpacifici, line 51: Why do you need to return this ?

py/sentry_streams/runner.py
```diff
@@ -32,56 +72,22 @@ def main() -> None:
...
     step_streams = {}
 
-    for source in p.sources:
+    for source in pipeline.sources:
         print(f"Apply source: {source.name}")
-        env_source = source.apply_source(env, environment_config)
+        env_source = translator.translate_step(source)
         step_streams[source.name] = env_source
-        iterate_edges(step_streams)
+        iterate_edges(step_streams, pipeline, translator)
```
fpacifici, line 90: Nit, what about passing the source to `iterate_edges` and let the function do the whole traversal rather than translate the source here?
If you pass the source and delegate to the `iterate_edges` method you do not have to initialize the mutable mapping `step_streams` here which makes it harder to understand the code.

Moreover the way in the current version you are translating steps in two places: one here and one inside `iterate_edges`. This adds cognitive overhead when trying to understand the algorithm.

py/sentry_streams/pipeline.py
```diff
@@ -0,0 +1,84 @@
...
+from sentry_streams.adapters.stream_adapter import StreamAdapter
+
+
+class RuntimeTranslator:
+    def __init__(self, runtime_adapter: StreamAdapter):
+        self.adapter = runtime_adapter
+
+    def translate_step(self, step: Step, stream: Optional[Any] = None) -> Any:
+        assert hasattr(step, "step_type")
+        next_step = step.step_type
```
fpacifici, line 16: `next_step` ?

py/sentry_streams/pipeline.py
```diff
@@ -0,0 +1,84 @@
...
+        translated_fn = getattr(self.adapter, next_step)
```
fpacifici, line 17: What about simplifying this for now and just do a bunch of instance of with exhaustive type checking so no new type can be missed?

```
from typing import assert_never

if isinstance(step, Sink):
        transalte sink
    elif isinstance(step, KafkaSource):
        translate source
    else:
        assert_never(widget.datapoint) # This fails if a new class is added that is not part of this if then else
```

py/tests/test_pipeline.py
```diff
@@ -0,0 +1,59 @@
...
+
+    libs_path = os.path.join("/".join(dir_path.split("/")[:-2]), "flink_libs")
+    assert libs_path is not None, "FLINK_LIBS environment variable is not set"
+
+    jar_file = os.path.join(os.path.abspath(libs_path), "flink-connector-kafka-3.4.0-1.20.jar")
+    kafka_jar_file = os.path.join(os.path.abspath(libs_path), "kafka-clients-3.4.0.jar")
+
+    env = StreamExecutionEnvironment.get_execution_environment()
+    env.set_parallelism(1)
+    env.add_jars(f"file://{jar_file}", f"file://{kafka_jar_file}")
```
fpacifici, line 35: You should not need all this part to test the translation.

py/tests/test_pipeline.py
```diff
@@ -0,0 +1,59 @@
...
+    step_streams = {}
+
+    for source in pipeline.sources:
+        print(f"Apply source: {source.name}")
+        env_source = translator.translate_step(source)
+        assert type(env_source) is DataStream
+        step_streams[source.name] = env_source
+        output_stream = iterate_edges(step_streams, pipeline, translator)
+
+    assert type(output_stream) is DataStreamSink
```
fpacifici, line 59: Consider doing assertions on the result of the `get_execution_plan()` method. It produces a representation of the Flink datastream in json. It looks like this 

```
{
  "nodes" : [ {
    "id" : 1,
    "type" : "Source: Custom Source",
    "pact" : "Data Source",
    "contents" : "Source: Custom Source",
    "parallelism" : 1
  }, {
    "id" : 2,
    "type" : "Sink: Print to Std. Out",
    "pact" : "Data Sink",
    "contents" : "Sink: Print to Std. Out",
    "parallelism" : 1,
    "predecessors" : [ {
      "id" : 1,
      "ship_strategy" : "FORWARD",
      "side" : "second"
    } ]
  } ]
}

```

fpacifici (approved):
Please address the comments before merging

py/sentry_streams/runner.py
ayirr7: There's kind of a small reason for returning it, I wanted to assert types of the input and output stream in test_pipeline. In order to do that and in order to be able to import `iterate_edges` I would need to return the output stream. I agree that it has no use outside of tests. 

I guess I can bake the assertion into the function `iterate_edges` itself, but it also seemed like a strange option considering the fact that the runner is then basically the test.

py/sentry_streams/flink/flink_adapter.py
```diff
@@ -0,0 +1,52 @@
...
+        kafka_consumer = FlinkKafkaConsumer(
+            topics=self.environment_config["topics"][topic],
+            deserialization_schema=SimpleStringSchema(),
+            properties={
+                "bootstrap.servers": self.environment_config["broker"],
+                "group.id": "python-flink-consumer",
+            },
+        )
+
+        return self.env.add_source(kafka_consumer)
```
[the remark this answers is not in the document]
ayirr7: As discussed offline, we need to return it because the StreamExecutionEnvironment doesn't have any reference to the DataStream being built.

py/sentry_streams/pipeline.py
```diff
@@ -0,0 +1,84 @@
...
+        step_config = {}
+
+        # build step-specific config
+        if hasattr(step, "logical_topic"):
+            step_config["topic"] = step.logical_topic
+
+        if stream:
+            return translated_fn(step_config, stream)
+        else:
+            return translated_fn(step_config)
```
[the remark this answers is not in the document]
ayirr7: That's way better, yeah

py/sentry_streams/runner.py
ayirr7: Also true. I think I did it this way to maintain the existing state of the code but there was no real reason for it to be structured this way.

py/sentry_streams/runner.py
ayirr7: fixed

py/tests/test_pipeline.py
ayirr7: you need the jar files to setup the StreamExecutionEnvironment

py/tests/test_pipeline.py
ayirr7: this is nice, yeah

py/sentry_streams/pipeline.py
ayirr7: In order to use `assert_never` I have to do some thing like or some combo of:

```
if isinstance(step, KafkaSource):
            return self.adapter.source(step)

        elif isinstance(step, Sink):
            return self.adapter.sink(step, stream)

        elif isinstance(step, WithInput):
            pass

        elif isinstance(step, KafkaSink):
            pass

        elif isinstance(step, Step):
            pass
        else:
            assert_never(step)
```

Perhaps I can just use a RuntimeError for now?

ayirr7: Also, a bigger point is that if we avoid using a static property on the dataclass to denote the step type and instead extend base classes (which generally makes more sense), then we end up with empty classes like this https://github.com/getsentry/streams/pull/8/files#diff-6e974549d7fef65165b6d06969fe9017e604a654169f1510052e5adce52fd135R59

There will likely be a few different Sinks, but Sinks are not much more than a `WithInput`.

fpacifici: The empty class is actually a "pattern" called marker interface even if not very common in python https://en.wikipedia.org/wiki/Marker_interface_pattern

Though we can say that is generally a symptom of a system that is not properly modularized. At this point I would not stress out too much about this as we are likely going to make fundamental changes to this data structure very soon and being able to unblock parallel work is more important.

fpacifici: Let's do this for this PR:
- create an enum with the types of steps we support (so the property is not a bare string that we can break)
- add the static property that contains a value of the enum above. 
- in the `translate_step` method enumerate the values of the enum and find the method to call that way. While a bit brittle at least we will be able to statically type check. Then we can make it smarter.
- you can use the `assert_never` to ensure the enum is checked exhaustively. 

Then let's add a few more primitives and then revisit the data structure.

PR #9, opened 2025-02-10
Add Map primitive

ayirr7:
Adds a basic 1:1 Map function

If we want to support 1:N map, I can add support for flatMap as well and hide both options (1:1 and 1:N) under our Map primitive -- let me know


py/sentry_streams/flink/flink_adapter.py
```diff
@@ -53,3 +57,16 @@ def sink(self, step: Step, stream: Any) -> Any:
         )
 
         return stream.sink_to(sink)
+
+    def map(self, step: Step, stream: Any) -> Any:
```
ayirr7, line 61: This currently assumes a specific format from users for how they define the path to their Map function

Will add checks for this

py/sentry_streams/flink/flink_adapter.py
```diff
@@ -53,3 +57,22 @@ def sink(self, step: Step, stream: Any) -> Any:
...
+            module = get_module(mod)
+
+        except ImportError as e:
+            raise e
+
+        imported_cls = getattr(module, cls)
+        imported_fn = getattr(imported_cls, fn)
+
+        # The output type must be specified
+        # TODO: Remove hardcoded output type
```
fpacifici, line 77: Are you sure it has to be specified ? 
I created a streaming app that does not specify that 

https://github.com/getsentry/streaming-tests/blob/main/pyflink_perf/flink_perf.py#L92-L103

py/sentry_streams/example_config.py
```diff
@@ -9,9 +9,16 @@
     logical_topic="logical-events",
 )
 
+map = Map(
+    name="mymap",
+    ctx=pipeline,
+    inputs=[source],
+    function="sentry_streams.sample_function.EventsPipelineMapFunction.simple_map",
```
fpacifici, line 16: Is it intentional to make this a bare string and not support a reference to a function ? 
Supporting references to a Callable:
Pros:
- type checking that the path is a valid function
Cons:
- That function has to be imported in this file which encourages people to write the function here.

py/sentry_streams/flink/flink_adapter.py
fpacifici: Not sure I get the issue. Are you talking about validating that the path to the function is actually a valid one ?
Isn't the check what you are doing at line 70 ?

py/sentry_streams/utils.py
fpacifici, line 1: Please let's not have a utils class.  Soon all the logic of the system will end up in this catch all module.

If this is meant to manage module names, please call it modules.py

py/tests/test_pipeline.py
fpacifici, line 1: Let's now have some proper unit tests. 

For this PR I would suggest:
1. A unit test for the map method on the Flink Adapter. Up to you if you want to create a real stream object, if you can spin up an environment quickly or whether to mock it.
2. Let's start a parametrized test for the translator.

If we keep piling up on this test we will miss the opportunity to test all corner cases with unit test and troubleshooting test failures will become very hard. Basically most tests should be simple and quick unit tests.

py/sentry_streams/pipeline.py
```diff
@@ -87,3 +88,13 @@ class KafkaSink(WithInput):
...
     step_type: StepType = StepType.SINK
+
+
+@dataclass
+class Map(WithInput):
+    """
+    A simple 1:1 Map, taking a single input to single output
+    """
+
+    function: str
```
fpacifici, line 99: Did we ever  consider how to provide the product access to the context of the pipeline? If yes is it going to be part of a follow up PR? If yes please add a todo.

fpacifici (changes requested):
Thanks, a few comments in line.
Do you plan to spend more time figuring out typing in a separate PR? If yes that's alright but please add a TODO

py/sentry_streams/flink/flink_adapter.py
ayirr7: So the default return type of Map is pickled byte array https://nightlies.apache.org/flink/flink-docs-release-1.18/api/python/reference/pyflink.datastream/api/pyflink.datastream.data_stream.DataStream.map.html?highlight=map#pyflink.datastream.data_stream.DataStream.map 

The output type for Map must be specified in cases where there is a next step after which does some sort of serialization. In this case, the next step is a Sink which uses the `SimpleStringSchema` to serialize strings into bytes objects for Kafka. 

Without the output type here, the job execution fails because Flink is not able to recognize or convert the output of Map as a Java String (got a bunch of errors about failing to cast to String). 

Anyways, I'll add this as a comment elsewhere

py/sentry_streams/flink/flink_adapter.py
ayirr7: Outdated comment, sorry

py/sentry_streams/utils.py
ayirr7: done!

py/sentry_streams/example_config.py
ayirr7: Left a comment to get back to this

py/sentry_streams/pipeline.py
ayirr7: Left TODO

py/sentry_streams/adapters/stream_adapter.py
```diff
@@ -19,6 +19,10 @@ def source(self, step: Step) -> Any:
     def sink(self, step: Step, stream: Any) -> Any:
         raise NotImplementedError
 
+    @abstractmethod
+    def map(self, step: Step, stream: Any) -> Any:
+        raise NotImplementedError
```
fpacifici, line 24: I think we can be more specific with the Step type.

As each step type is translated with its own specific method we can make the adapter take `step: Map` and assert the type is right in the RuntimeTranslator:
```
elif step_type is StepType.MAP:
            return self.adapter.map(step, stream)
```

The rationale for this is that we can keep the logic that manages dynamic types in the same place at least.
In the current architecture the translator cannot request a specific type of Step by definition, so it has to infer, from the type, the method to call. 
Though we can still keep the adapter strictly typed as each method only takes one type of step. This allows us to rely on assertions only in one place and take advantage of typing as much as we can.

py/sentry_streams/flink/flink_adapter.py
```diff
@@ -53,3 +63,21 @@ def sink(self, step: Step, stream: Any) -> Any:
...
+
+        assert hasattr(step, "function")
+        fn_path = step.function
+        mod, cls, fn = fn_path.rsplit(".", 2)
+
+        try:
+            module = get_module(mod)
+
+        except ImportError as e:
+            raise e
```
fpacifici, line 77: nit
```
except ImportError:
    raise
```

py/sentry_streams/utils.py
fpacifici, line 1: please remove.

py/tests/test_pipeline.py
```diff
@@ -78,4 +86,121 @@ def test_pipeline() -> None:
...
+
+
+@pytest.mark.parametrize("pipeline,expected_plan", [basic(), basic_map()])
+def test_pipeline(
+    setup_basic_flink_env: tuple[StreamExecutionEnvironment, RuntimeTranslator],
+    pipeline: Pipeline,
+    expected_plan: MutableMapping[str, list[dict[str, Any]]],
+) -> None:
+    env: StreamExecutionEnvironment
+    translator: RuntimeTranslator
```
fpacifici, line 161: You should not need to specify the type here.

fpacifici (approved):
please address the comments before merging.

PR #18, opened 2025-02-20
WIP; windowing, aggregation

ayirr7:


py/sentry_streams/pipeline.py
```diff
@@ -114,5 +126,21 @@ class Map(WithInput):
...
+    # group_by_key: refactor to Callable reference
+    group_by_key: Callable[..., Any]
+    # windowing mechanism, is this going to be mandatory?
+    # windowing: Window
+    # aggregation (use standard accumulator)
+    aggregate_fn: Accumulator
+    step_type: StepType = StepType.REDUCE
+    # storage: a fixed (enum?) set of storage backends we provide
+    # consider making this a class
+    storage: StateBackend = StateBackend.HASH_MAP
```
fpacifici, line 144: Isn't the storage hidden by the accumulator at times ?
Think about these cases:
- Aggregation with windowing and producing the output. Here we can separate aggregaiton function from store as we can store the aggregation state where we want.
- Global state: accumulating counts in postgres. Here the storaeg is application specific so there is no point in separating the two.

py/sentry_streams/user_functions/agg_template.py
```diff
@@ -0,0 +1,28 @@
...
+    def add(self, acc: Any, value: Any) -> Any:
+        raise NotImplementedError
+
+    @abstractmethod
+    def get_output(self, acc: Any) -> Any:
+        raise NotImplementedError
+
+    @abstractmethod
+    def merge(self, acc1: Any, acc2: Any) -> Any:
+        raise NotImplementedError
```
fpacifici, line 21: Hint. Try to make this generic rather than have `Any` as type everywhere.

py/sentry_streams/user_functions/agg_template.py
```diff
@@ -0,0 +1,28 @@
+from abc import ABC, abstractmethod
+from typing import Any
+
+
+class Accumulator(ABC):
```
fpacifici, line 5: Do you have a sense on which accumulators we will provide out of the box ?

fpacifici (commented):
Some early comments

py/sentry_streams/pipeline.py
```diff
@@ -114,5 +126,21 @@ class Map(WithInput):
...
+    function: Callable[..., Any]
     step_type: StepType = StepType.MAP
+
+
+@dataclass
+class Reduce(WithInput):
+    # group_by_key: refactor to Callable reference
+    group_by_key: Callable[..., Any]
+    # windowing mechanism, is this going to be mandatory?
+    # windowing: Window
```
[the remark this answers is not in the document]
ayirr7: Yep, plan was to add that next

py/sentry_streams/pipeline.py
```diff
@@ -114,5 +126,21 @@ class Map(WithInput):
...
-    function: str
+    function: Callable[..., Any]
     step_type: StepType = StepType.MAP
+
+
+@dataclass
+class Reduce(WithInput):
+    # group_by_key: refactor to Callable reference
+    group_by_key: Callable[..., Any]
+    # windowing mechanism, is this going to be mandatory?
```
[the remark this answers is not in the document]
ayirr7: as discussed offline, we'll support that use case (for now) using windowed aggregation + sink. 

So windowing will be required in this iteration of the API

py/sentry_streams/pipeline.py
ayirr7: I _think_ this is the same discussion we had related to https://github.com/getsentry/streams/pull/18#discussion_r1964903269

I can leave a TODO to come back to this in a next PR since this once is getting large

py/sentry_streams/user_functions/agg_template.py
ayirr7: will do, right now just getting things working with Flink

ayirr7:
Still todo for this PR:
- Address typing comment above (move to Generics)
- Add tests

py/sentry_streams/pipeline.py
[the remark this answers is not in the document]
ayirr7: Added by allowing users to configure windows and triggers

py/sentry_streams/flink/flink_fn_translator.py
```diff
@@ -0,0 +1,114 @@
...
+    Convert a timedelta object to a Time type representation in milliseconds.
+    """
+
+    total_milliseconds: int = int(timestamp.total_seconds() * 1000)
+
+    flink_time: Time = Time.milliseconds(total_milliseconds)
+    return flink_time
+
+
+class FlinkWindows:
```
ayirr7, line 74: I think this translation needs some work. 

Right now, the API exposes only a few specific window, trigger combinations that a user can pick from. One benefit of this is that we can just use some of the `WindowAssigners` that Flink offers out of the box. Down the road though, this translation will need to be more flexible. 

Flink exposes a `WindowAssigner` interface which we could try to implement as a translation mechanism. This is a possibility, will just require a little more understanding of Flink's internals. 

Open to any other ideas as well.

py/sentry_streams/user_functions/agg_template.py
ayirr7: Not yet. I think that might be a good idea for next PR. Leaving a TODO

py/sentry_streams/pipeline.py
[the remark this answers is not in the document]
ayirr7: Another point. Windows that close by event time are inherently tied to the watermarking strategy. 

This strategy can be set directly on the source especially if we're using something like Kafka. We can also allow users to configure their own as part of a step like Reduce, though I believe that's not recommended because that would mean a new watermarked stream is being produced as part of the step. 

WDYT about how we should configure watermarking strategy?

py/sentry_streams/flink/flink_adapter.py
```diff
@@ -81,5 +91,45 @@ def map(self, step: Map, stream: Any) -> Any:
...
+
+        else:
+            windowed_stream = time_stream.window_all(window_assigner).trigger(trigger)
+
+        return_type = agg.get_output.__annotations__["return"]
+
+        # TODO: Figure out a systematic way to convert types
+        return windowed_stream.aggregate(
+            FlinkAggregate(agg),
+            accumulator_type=Types.TUPLE([Types.STRING(), Types.INT()]),
```
markstory, line 133: Could accumulators collect other aggregation types as well?

py/sentry_streams/pipeline.py
markstory: > Global state: accumulating counts in postgres. Here the storaeg is application specific so there is no point in separating the two.

We could separate the accumulation logic from the flush to postgres concerns. We may also have scenarios where we need to flush streaming aggregations to redis.

py/sentry_streams/user_functions/function_template.py
```diff
@@ -0,0 +1,38 @@
...
+    Define these functions to build a custom
+    Accumulator for aggregation.
+    """
+
+    @abstractmethod
+    def create(self) -> Any:
+        raise NotImplementedError
+
+    @abstractmethod
+    def add(self, acc: Any, value: Any) -> Any:
```
untitaker, line 17: I would try to type this out properly soon. from what I understand `value` comes from the previous step, and `acc` can be a custom type specific to the reduce step.

if the two types are equal (like in your word counter example), maybe we can allow the user to only specify one function, since `add` and `merge` would do the same thing?

py/sentry_streams/example_config.py
```diff
@@ -13,12 +17,26 @@
...
+# Windows are assigned 3 elements.
+# But aggregation can be triggered as soon as 2 elements are seen.
+reduce_window = TumblingCountWindow(trigger=CountingTrigger(2), window_size=3)
+
+reduce = Reduce(
+    name="myreduce",
+    ctx=pipeline,
+    inputs=[map],
+    windowing=reduce_window,
+    aggregate_fn=WordCounter(),
```
untitaker, line 33: WordCounter is usercode, shouldn't we make it so that the pipeline definition does not directly import user code? same question about map

fpacifici: re: pipeline definition importing user code. 

I don't think that's a problem as long as the imported user code does not impact the semantics of the pipeline definition itself. 

Example of ok behavior:
- user code contains the map function, the pipeline definition imports that and references it inside the python file that adds the Map step. This is good ergonomics for the product engineers that do not have to split the two in a way we cannot enforce anyway. 

Example of bad behavior:
- Import user code, add a Map step to the pipeline conditionally according to an `if` branch which relies on user code. That defies the purpose of a declarative and static pipeline definition.

So we need to dig a bit deeper on how to enforce the proper behavior and forbid the negative one. I suspect preventing user code where the pipeline definition entirely is not going to solve the problem (how would we enforce that?). If we are able to enforce that separation we should also be able to prevent the bad behavior without preventing the user from referencing user code from the python module where the pipeline is.

untitaker: >  user code contains the map function, the pipeline definition imports that and references it inside the python file that adds the Map step. This is good ergonomics for the product engineers that do not have to split the two for pipelines that are entirely defined in the same codebase.

if we allow user code to be referenced by function pointer as it is done here, then it's not possible to load the graph and reason about it without importing all of sentry. we should find a way to allow people to reference user code without forcing the pipeline builder to import it. in previous examples this meant referencing functions by import path rather than pointer.

fpacifici: >  we should find a way to allow people to reference user code without forcing the pipeline builder to import it

What is your concern? We cannot do much about the pipeline without importing the application code anyway.

untitaker: isn't that one of the things we agreed on inside of that meeting, that we should not allow people to import `django.conf` etc? this is what it means to me to import the application. we should draw a clear line between the pipeline definition and the application code that can and will do arbitrary things at import time.

fpacifici: > isn't that one of the things we agreed on in-person, that we should not allow people to import django.conf etc?

I would not be too strict about those early decisions. Now that we are seeing the real code and product engineers have a chance to try to build build on it we should be flexible.
For example there is already feedback against defining the pipeline with classes that reference the previous step: https://www.notion.so/sentry/Response-to-Filippo-s-Comments-1978b10e4b5d80dbb4c8fcd0ca687753?pvs=4#19a8b10e4b5d805bbbcde02a9192856b.

> that we should not allow people to import django.conf etc? this is what it means to me to import the application. we should draw a clear line between the pipeline definition and the application code that can and will do arbitrary things at import time.

I see your concern, though we will have to import all of that to run the pipeline anyway. So, since preventing application code import will be a controversial decision (pipeline authors will push back) and will be hard to enforce, we need to be very clear on the rationale and the tradeoff. 
What about this: let's capture pros and cons here as it has impact on this specific point then I will capture it in our architecture docs and bring the conversation out.

> that issue would be exarcerbated once we get into cross-application pipeline definitions, where with this design, in order to figure out the pipeline graph we'd then have to import both sentry and snuba.

You are right on this. I put some thoughts into this lately and I believe in the end we will have to support a yaml style representation for pipelines that that are cross code base. That will always be supported and likely the priamry way to define a pipeline. 
I see the python one as possible syntactic sugar to provide a better ergonomics and reduce friction for the pipelines that are fully contained in the python code base. The semantics of the produced graph would be identical.

py/sentry_streams/flink/flink_adapter.py
ayirr7: For sure. We need to figure out a way to sytematically convert Python types into types that Flink understands. 

Hence the TODO comment

py/sentry_streams/window.py
```diff
@@ -0,0 +1,108 @@
...
+
+@dataclass
+class CountingTrigger(Trigger):
+    """
+    Specifically triggers window functions
+    (for example, aggregates) when the number
+    of elements in a window exceeds the count.
+    """
+
+    count: int
```
fpacifici, line 33: Isn't the type of trigger coupled to the type you have in the Window (datetime vs count)?
In other words, I'd assume we cannot use an IntervalTrigger in a SlidingCountWindow.

If I am correct, I think we should force a specific type of window to have a specific type of trigger.
Can we do this with type hints maybe making the Trigger generic with respect to the type used inside and force a window adn trigger to use the same type when the type hint is resolved?

py/sentry_streams/user_functions/function_template.py
```diff
@@ -0,0 +1,38 @@
+from abc import ABC, abstractmethod
+from typing import Any
+
+
+class Accumulator(ABC):
```
fpacifici, line 5: How is this going to be related with the storage choice? 
Are Accumulator and Storage separate entities or would they be connected ?

py/sentry_streams/flink/flink_fn_translator.py
fpacifici: Exposing a subset of the primitives is ok. Actually it is a desirable behavior for a platform:
- We can be opinionated on the data model and the capabilities we want to expose rather than simply inheriting everything the runtime underneath supports. This allows us to focus on what is important for the application and allows for maintainability.
- We have an easier time in supporting multiple runtimes.
- The value of the platform is often to put the right restrictions to guarantee we follow an architecture, which allows for consistency in how we solve problems.

Obviously finding the right constraints and the right primitives to expose is the hard part. At this point what you provided is more than enough.

py/sentry_streams/window.py
ayirr7: I think there are cases in which the two don't have to match, for example:
- Global windows. These will window all data seen if they have the same key. You can apply a variety of Triggers to this. 
- Your condition for closing windows is different from the condition for triggering a computation. What if you want to window data by their event times (and you could have late data), but you want to trigger computation as soon as you have X events in the window? (Something like an alert threshold being exceeded?)

Flink also allows you to plug in/customize triggers with whatever windowing mechanism you have. 

Don't know if we want to anticipate this level of flexibility already. WDYT?

py/sentry_streams/example_config.py
fpacifici: I created a daci for the discussion https://www.notion.so/sentry/DACI-How-do-we-wire-operators-with-each-other-in-the-pipeline-1a58b10e4b5d80ff8f73f55b7feb4405

py/sentry_streams/window.py
ayirr7: Discussed offline:

since we don't have any use cases in mind which will need a trigger that's different from the window type, we can just remove triggers altogether for now.

ayirr7:
Reminder to self: 

- Think about how to provide input and output types for each step in the Pipeline, and how to represent that
- How does that correlate with translation to runtime-specific Types/abstractions

fpacifici:
> Think about how to provide input and output types for each step in the Pipeline, and how to represent that

We have a task for this https://github.com/getsentry/streams/issues/25

py/sentry_streams/pipeline.py
```diff
@@ -145,3 +159,12 @@ class Filter(TransformStep[bool]):
 
     function: Union[Callable[..., bool], str]
     step_type: StepType = StepType.FILTER
+
+
+@dataclass
+class Reduce(WithInput, Generic[MeasurementUnit, InputType, IntermediateType, OutputType]):
+    windowing: Window[MeasurementUnit]
+    aggregate_fn: Accumulator[InputType, IntermediateType, OutputType]
+    group_by_key: Optional[GroupBy] = None
```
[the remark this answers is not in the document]
ayirr7: Turns out, there's a lot of flexibility with Flink. Grouping by on a stream does **not necessarily** produce a (K, V) pair output stream (you could have the same type in the input and keyed stream even). With Flink, the intermediate / accumulator type doesn't have to be a (K, V) either.

You could do something like this in Flink:

```
class Event:
   
     key: ...
     count: ...
     sum: ...


class AverageCalculator(Accumulator[Event, RiyasAverage, Tuple[RiyasAverage, ...]]):

    def create(self) -> RiyasAverage:
        return RiyasAverage(...)

    def add(self, acc: RiyasAverage, value: Event) -> RiyasAverage:
        ...

    def get_output(self, acc: RiyasAverage) -> Tuple[RiyasAverage, ...]:
       ....

    def merge(self, acc1: RiyasAverage, acc2: RiyasAverage) -> RiyasAverage:
        ...


output_stream = input_stream.group_by(lambda event: event.key).window(...).aggregate(AverageCalculator)

```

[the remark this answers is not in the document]
ayirr7: > the output would not contain the specific window (group by key)

What if the aggregation doesn't produce a result stream but instead only flushes windows to a DB (for example). Do we want to allow this? 

> the output would not contain the specific window (group by key)

Group by is optional right now. We'll probably have a `Union[]` of the non-group by and the group-by Accumulator types.

py/sentry_streams/pipeline.py
ayirr7: I've taken a stab at this via the `AggregationBackend`. Right now, the backend can take the output of the Accumulator and flush it to a configurable storage. 

Let me know what you all think of this interface.

py/sentry_streams/user_functions/function_template.py
ayirr7: Added `AggregationBackend`. It's a WIP that needs some feedback.

ayirr7:
I will add some unit tests next. 

Mainly, I'd like to get some feedback on the state of the interfaces now.

py/sentry_streams/runner.py
```diff
@@ -47,7 +53,7 @@ def iterate_edges(p_graph: Pipeline, translator: RuntimeTranslator) -> None:
                     else:
                         next_step: WithInput = cast(WithInput, p_graph.steps[output_step_name])
                         print(f"Apply step: {next_step.name}")
-                        next_step_stream = translator.translate_step(next_step, input_stream)
+                        next_step_stream = translator.translate_step(next_step, input_stream)  # type: ignore
```
ayirr7, line 56: The type: ignore here is unfortunate and I don't know how to get around it very easily just yet. 

The issue is, the `RuntimeTranslator` has the generic types `Stream, StreamSink`. And the `translate_step` function only works with `Optional[Stream]`. While the code never actually does `translate_step()` on a `StreamSink`, the type checker will complain about this. 

One way around it would be to support a generic `Stream` type that somehow encapsulates both Stream/Sink in the `StreamAdapter`. Thoughts?

py/sentry_streams/adapters/loader.py
```diff
@@ -1,14 +1,19 @@
...
 
 
 class AdapterType(Enum):
     FLINK = "flink"
 
 
-def load_adapter(adapter_type: Union[AdapterType, str], config: PipelineConfig) -> StreamAdapter:
+def load_adapter(
+    adapter_type: Union[AdapterType, str], config: PipelineConfig
+) -> Union[FlinkAdapter]:
```
ayirr7, line 16: Also unfortunate. I can't get it to work unless I do this or return `StreamAdapter[...Flink specific types...]`.

fpacifici (commented):
I have an idea to ensure we feel good with the API. Please add an `examples` package inside the sentry_streams package. There we can have one file per example.
Let's have a few examples for reduce.  They will be bare bones example here but they should aim at addressing realistic use cases.
Some of them will be repetitive at first because they will differentiate themselves through some primitives we do not yet have (like storage type).
Let's build these:
- The alerts system. 
- - You ingest events with some numeric measurement like latency, a timestamp and a project it.
- - You want to produce two time series: 
- - - a count of events over a sliding window of 5 mintues per project id. THis is produced every minute
- - - a p95 (pretend you have a state that is able to manage this, you can mock it) over 5 mintues every minute per project.
- A "totally hypothetical" trace buffer.
- - ingest spans: they have a span id a trace id and a duration and a timestamp.
- - accumulate spans for a trace for 5 minutes max. Once the window is over the spans you accumulated is a "segment". compute the total duration of the segment, produce a segment message that contains the total duration and the list of spans.
- - if a span called "end" shows up, flush the segment early.
- The batching consumer
- - Let's pretend we had a database where we need to write in batches. Who would ever do that ?
- - You get events one by one, you accumulate them into batches in memory, when a batch time or size is reached you flush the entire batch to a step that does something with it. 
- The totally hypothetical billing consumer 
- - Receive outcomes data 
- - Filter out the outcomes we do not care for 
- - Batch outcomes into this hypothetical data structure: [_PendingBuffer](https://github.com/getsentry/getsentry/blob/master/getsentry/consumers/outcomes_consumer.py#L77)
- - - This data structure builds a hashmap in memory where the key is a number of filters on the outcome. Let's say state and data category. and the value is the count of outcomes that correspond to those filters.
- - After a max size and time flush which means write the hashmap into a hypothetical store called redis. 

You will certainly miss some primitives for those. The goal is not to implement all of them in this PR but to identify them and ensure the current interfaces are going in the right direction. So feel free to mock out what is missing.

py/sentry_streams/flink/flink_fn_translator.py
```diff
@@ -0,0 +1,133 @@
...
+
+    def get_result(self, accumulator: IntermediateType) -> OutputType:
+        return self.acc.get_output(accumulator)
+
+    def merge(
+        self,
+        acc_a: IntermediateType,
+        acc_b: IntermediateType,
+    ) -> IntermediateType:
+        return self.acc.merge(acc_a, acc_b)
```
fpacifici, line 75: I am not sure I actually get the goal of this class.
This class is purely a wrapper around the `Accumulator` you pass in the initializer. It does not seem to add anything. 
- Every method call just routes the same call to the Accumulator class without adding any additional parameter. It seems to me the `Accumulator[InputType, IntermediateType, OutputType]` class should fit everywhere you use this class without adding this abstraction. 
- Am I missing anything ?

Is it here only to decouple Flink from our API further.

py/sentry_streams/flink/flink_fn_translator.py
```diff
@@ -0,0 +1,133 @@
...
+    """
+    Takes the Streams API's GroupBy and provides a thin
+    wrapper to convert to Flink's GroupBy mechanism.
+    """
+
+    def __init__(self, group_by: GroupBy) -> None:
+        self.group_by = group_by
+
+    def get_key(self, value: Any) -> Any:
+        return self.group_by.get_group_by_key(value)
```
fpacifici, line 88: Similar comment I added to the previous class. Is this class needed to map the `get_key` function call needed by flink to `get_group_by_key` ?

py/sentry_streams/flink/flink_fn_translator.py
```diff
@@ -0,0 +1,133 @@
...
+    Convert a timedelta object to a Time type representation in milliseconds.
+    """
+
+    total_milliseconds: int = int(timestamp.total_seconds() * 1000)
+
+    flink_time: Time = Time.milliseconds(total_milliseconds)
+    return flink_time
+
+
+class FlinkWindows(Generic[MeasurementUnit]):
```
fpacifici, line 102: I don't think you need to make this class generic. 
You should be able to just make 
```
self.window: Window[MeasurementUnit] = window
```

This is because, while `Window` is a generic class. This class never uses the object `Window` is generic on. See no method in this class uses or returns `MeasurementUnit`, which means this class may not be parametric with respect to `MeasurementUnit`. 

Though looking at the class further I don't think you need a class at all. I suspect there will only be one calls to `build_window` for each instance of this class. If you make it a function the result will be simpler as nothing is generic.

py/sentry_streams/runner.py
```diff
@@ -92,7 +98,7 @@ def main() -> None:
     }
 
     pipeline: Pipeline = pipeline_globals["pipeline"]
-    runtime = load_adapter(AdapterType.FLINK, environment_config)
+    runtime: FlinkAdapter = load_adapter(AdapterType.FLINK, environment_config)
```
fpacifici, line 101: Why does this need to know about `FlinkAdapter` ? I'd assume you added the type hint here because `RuntimeTranslator(runtime)` somehow needs to know the specific Adapter. But that sounds coutner to the design that abstracts the adapter away from the translation code.

py/sentry_streams/runner.py
fpacifici: Not sure I fully get this issue.
Is mypy complaining that `input_stream` is not a `Stream` ?
The fact that you get here with `input_stream` that could be `Stream` or `StreamSink` and rely on the fact that `translate_step` is never called on `StreamSink` points to something not sound on the types we sed around and maybe we should revisit this algorithm entirely. 

So first hint: remove the `type: ignore` and replace it with a `cast`, so the code declares exactly which type is problematic without disabling type checking entirely for the method. 

What happens if you type `step_streams = {}` and allow that to only contain Stream and StreamSource? Where is the type checking failing ?

py/sentry_streams/user_functions/function_template.py
```diff
@@ -0,0 +1,79 @@
...
+
+
+class AggregationBackend(ABC, Generic[OutputType]):
+    """
+    A storage backend that is meant to store windowed aggregates. Configurable
+    to the type of storage.
+    """
+
+    @abstractmethod
+    def flush_aggregate(self, aggregate: OutputType) -> None:
```
fpacifici, line 16: Is this actually supposed to be a class ? In other word, is it supposed to handle state ? Can it be a function?

py/sentry_streams/user_functions/function_template.py
```diff
@@ -0,0 +1,79 @@
...
+        Flush a windowed aggregate to storage. Takes in the output from
+        the Accumulator.
+        """
+
+
+class Accumulator(ABC, Generic[InputType, IntermediateType, OutputType]):
+    """
+    The standard Accumulator template.
+    Define these functions to build a custom
+    Accumulator for aggregation.
```
fpacifici, line 27: This will be the entry point for user to stateful streaming. 
Let's expand on this docstring explaining how it is supposed to be used and providing an example.
When you provide more examples for the API we will not need to have an example in the docstring.

py/sentry_streams/user_functions/function_template.py
```diff
@@ -0,0 +1,79 @@
...
+    """
+
+    def __init__(self, backend: Optional[AggregationBackend[OutputType]] = None):
+        self.backend = backend
```
fpacifici, line 31: The fact that no method in this class reference the backend is a hint that the backend attribute should not be defined at this level and should be left to the subclasses. 
If having a constraint defined at this level was semantically useful, then the methods below would use it.

Since the interface of the accumulator does nothing useful for the backend, is the backend actually needed as an abstraction? OR is it an implementation detail of specific implementation of the accumulator ?

py/sentry_streams/user_functions/function_template.py
```diff
@@ -0,0 +1,79 @@
...
+    @abstractmethod
+    def add(self, acc: IntermediateType, value: InputType) -> IntermediateType:
+        """
+        Add values to the Accumulator. Can produce a new type which is different
+        from the input type.
+        """
+        raise NotImplementedError
+
+    @abstractmethod
+    def get_output(self, acc: IntermediateType) -> OutputType:
```
fpacifici, line 49: What about `get_value` ? 
It seems more in line with the semantics of an accumulator which is a system that accumulates data into a state and that can return the state at any moment. 
get_output seems to imply there is some sort of computation done by the accumulator, which may not be the case.

This may sound pedantic but I believe that, when defining an api, naming things in a way that precisely identify the semantics pays back later when people use it.

py/sentry_streams/user_functions/sample_agg.py
```diff
@@ -0,0 +1,39 @@
+from typing import MutableMapping
+
+from sentry_streams.user_functions.function_template import (
+    Accumulator,
+    AggregationBackend,
+)
+
+WordCountTuple = tuple[str, int]
+WordCountStr = str
```
fpacifici, line 9: Inline these ? You do not need to declare the type when instantiating a class that is generic. 
Defining the type aliases is generally useful when the name adds something semantically to the bare types. Here it does not seem it is happening as the name of the alias is basically the name of the bare types.

py/sentry_streams/user_functions/sample_group_by.py
fpacifici, line 1: Please let's keep the examples separate from the classes that are actually part of the API. Let's create an `examples` module at the root of sentry_streams

py/sentry_streams/window.py
```diff
@@ -0,0 +1,66 @@
...
+    Each Window can have a trigger plugged in.
+    """
+
+
+@dataclass
+class SlidingCountWindow(Window[MeasurementUnit]):
+    """
+    A sliding window which is configured
+    by counts. Size and slide are both
+    in terms of number of elements.
```
fpacifici, line 21: Nit: you can make these lines a little longer

py/sentry_streams/window.py
```diff
@@ -0,0 +1,66 @@
...
+@dataclass
+class Window(Generic[MeasurementUnit]):
+    """
+    A generic representation of a Window.
+    Each Window can have a trigger plugged in.
+    """
+
+
+@dataclass
+class SlidingCountWindow(Window[MeasurementUnit]):
```
fpacifici, line 17: I think I still see the original problem here: two dimensions (whether the window is sligin or tumbling) and two ways to measure the window (counter and event) -> 4 classes. This multiplicative explosions of the classes is problematic: every time you add a window type or a way to measure the window, the number of classes to add is multiplicative and some will be missed.

What I meant is to use inheritance to represent window types (as the  attributes are legit different between the window types) and use the generic for the measurement type which only impacts the type of the parameters above. 
This means you do not have to create a new class for each combination as the user can simply create an instance of the window providing parameters:
```
SlidingWindow(
   window_size = 10
   window_slide = 3
)
```

Is enough.

Did you create each class at runtime in order to be able to identify the right class for flink ? https://github.com/getsentry/streams/pull/18/files#diff-b23be9ade09256acec424f43ee4a0bde80ff1d4271a5e266fcbfe0d508619ca9R118-R125

IF yes you can rely on the type of the window_size at runtime as it is statically forced to be one of the two.

fpacifici (commented):
I think it is getting close. Thanks
A few comments in line and then let's see how it looks like with the examples.

py/sentry_streams/adapters/stream_adapter.py
```diff
@@ -1,12 +1,31 @@
...
 
 PipelineConfig = Mapping[str, Any]
 
 
-class StreamAdapter(ABC):
+Stream = TypeVar("Stream")
+StreamSink = TypeVar("StreamSink")
+
+
+class StreamAdapter(ABC, Generic[Stream, StreamSink]):
```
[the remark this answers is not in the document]
ayirr7: Not needed for the sake of this PR. 

I received a comment or two on it to try to clean up typing in general in this PR. This was my attempt at that.

py/sentry_streams/flink/flink_adapter.py
```diff
@@ -114,15 +141,62 @@ def sink(self, step: Step, stream: Any) -> Any:
...
+
+            windowed_stream: Union[WindowedStream, AllWindowedStream] = keyed_stream.window(
+                window_assigner
+            )
+
+        else:
+            windowed_stream = time_stream.window_all(window_assigner)
+
+        acc_type = agg.create.__annotations__["return"]
+        return_type = agg.get_output.__annotations__["return"]
```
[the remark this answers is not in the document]
ayirr7: It doesn't work for now, but maybe we can enforce type hints? 

Do you have other ideas for this?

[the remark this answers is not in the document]
fpacifici: You can require that the `aggregate_fn` either provides type hints (this not returning empty) or has to specify the type we want inside the reduce Step as an attribute. The condition can be validated here.

py/sentry_streams/examples/billing.py
```diff
@@ -0,0 +1,108 @@
...
+
+
+def build_outcome(value: str) -> Outcome:
+
+    d: Outcome = json.loads(value)
+
+    return d
+
+
+class OutcomesBackend(AggregationBackend[PendingBuffer]):
```
fpacifici, line 27: Let's say this was a redis storage. 
We would want to control the place we write the buffer to Redis. The product should only know about the abstract data structures they are writing like keys and values.
What about we had a KVBackend (Key Value pair Backend).
We extend it to provide a Redis implementation and make PendingBuffer provide a method that returns the keys and values for the storage.

You do not need to implement it here unless it is quick. But let's track this development.

py/sentry_streams/examples/billing.py
```diff
@@ -0,0 +1,108 @@
...
+    Upon the closing of a window, the Buffer is flushed to a
+    sample backend (the OutcomesBackend). As of now this backend
+    is not a mocked DB, it is a simple hash map.
+    """
+
+    def __init__(self, backend: Optional[OutcomesBackend] = None):
+        self.backend = backend
+
+    def create(self) -> PendingBuffer:
+        buffer = PendingBuffer()
```
fpacifici, line 52: Is there any reason for PendingBuffer not being an `Accumulator` itself and adding this intermediate step ?
I think we may be able to skip the step. 

Imagine this 

```
class OutcomesBuffer(KeyValueAccumulator[Outcome]):
   def add(self, acc: PendingBuffer, value: Outcome) -> PendingBuffer:
        if "state" in outcome:
            self.map["state"] += 1
        elif "data_cat" in outcome:
            self.map["data_cat"] += 1
         .......

   def merge_with(self, other: OutcomesBuffer) -> Self:
         self.map.udpate(other.map)

   def get_output(self) -> Mapping[str, ...] :
        # Whatever business logic needed to export the keys
        return self.map
```

Then in the Reduce step:

```
reduce = Reduce(
    name="myreduce",
    ctx=pipeline,
    inputs=[map],
    windowing=reduce_window,
    aggregate_fn=KVAccumulator(OutcomesBuffer()) # To be configured for redis in the config file.
)
```

The idea is that the user only needs to tell us:
- How to add outcomes to the state
- How to merge states
- That the state is a KV pair state
- How to extract Keys and Values from the state to flush.

Am I missing anything here?

fpacifici (commented):
Thanks I think this approach makes reasoning about the api on real use cases easier.

py/sentry_streams/examples/billing.py
markstory: > How to extract Keys and Values from the state to flush.

We'd also need a step/stage that consumes the key/value pairs and flushes them to storage, but that could be an application logic filled step.

py/sentry_streams/examples/batching.py
```diff
@@ -0,0 +1,58 @@
...
+
+source = KafkaSource(
+    name="myinput",
+    ctx=pipeline,
+    logical_topic="logical-events",
+)
+
+# A sample window.
+# Windows are assigned 4 elements.
+reduce_window = TumblingWindow(window_size=4)
```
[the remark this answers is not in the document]
markstory: Having windows based on event time, and wall clock time are two additional windowing scenarios I can see us needing.

py/sentry_streams/examples/billing.py
markstory: Couldn't `storage_map` act as placeholder for redis, or another key value store?

py/sentry_streams/examples/billing.py
```diff
@@ -0,0 +1,108 @@
...
+
+    def __init__(self) -> None:
+        self.storage_map: dict[str, int] = {"state": 0, "data_cat": 0}
+
+    def flush_aggregate(self, aggregate: PendingBuffer) -> None:
+        """
+        Simply overrides the current values in the storage.
+        """
+        self.storage_map["state"] = aggregate.get_key("state")
+        self.storage_map["data_cat"] = aggregate.get_key("data_cat")
```
markstory, line 37: If the window aggregation is also flushing to 'persistent storage' what would be in the `reduce` stream? Would it be aggregates that were flushed?

py/sentry_streams/examples/batching.py
```diff
@@ -0,0 +1,58 @@
...
+    def get_output(self, acc: list[str]) -> str:
+        return " ".join(acc)
+
+    def merge(self, acc1: list[str], acc2: list[str]) -> list[str]:
+        return acc1 + acc2
+
+
+pipeline = Pipeline()
+
+source = KafkaSource(
```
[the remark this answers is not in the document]
ayirr7: Yes, that's what I've been doing -- perhaps I skipped it here. 

Can add

py/sentry_streams/examples/billing_buffer.py
```diff
@@ -0,0 +1,27 @@
...
+            self.map = {"state": 0, "data_cat": 0}
+
+    def add(self, outcome: Outcome) -> None:
+
+        if "state" in outcome:
+            self.map["state"] += 1
+
+        elif "data_cat" in outcome:
+            self.map["data_cat"] += 1
+
```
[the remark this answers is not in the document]
ayirr7: woops, misunderstood the example then

py/sentry_streams/flink/flink_fn_translator.py
ayirr7: Flink expects an `AggregateFunction` for the aggregation on windows. We, in the Streams API, want to provide a standard template for an Accumulator that users will implements. 

We will translate it to various runtime-specific interfaces. This is the Flink interface. 

Does that make sense? I also added a docstring to explain it.

py/sentry_streams/flink/flink_fn_translator.py
ayirr7: changed it to a function, it's cleaner, thanks

py/sentry_streams/examples/batching.py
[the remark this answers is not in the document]
ayirr7: Yes we have event time-based windowing.

Processing time-based windowing can also be added.

I will show examples of both.

py/sentry_streams/examples/batching.py
```diff
@@ -0,0 +1,58 @@
...
+    KafkaSink,
+    KafkaSource,
+    Pipeline,
+    Reduce,
+)
+from sentry_streams.user_functions.function_template import Accumulator
+from sentry_streams.window import TumblingWindow
+
+
+class BatchBuilder(Accumulator[str, list[str], str]):
```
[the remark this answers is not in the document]
ayirr7: I like the idea, I can try it

py/sentry_streams/examples/billing.py
ayirr7: Implemented, let me know what you all think

If we think it's not a good template as-is, I'll tackle in a separate PR

py/sentry_streams/examples/billing.py
ayirr7: Implemented a scaffolding for KVAccumulator and KVBackend

py/sentry_streams/flink/flink_fn_translator.py
ayirr7: Yes, see https://github.com/getsentry/streams/pull/18#discussion_r1980423249

py/sentry_streams/examples/batching.py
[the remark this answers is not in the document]
ayirr7: The SpansBuffer uses event time-based windowing.

I can add processing time-based windowing in another PR when we have an apt example for it.

py/sentry_streams/user_functions/function_template.py
ayirr7: It's been updated a bit. LMK what you think

py/sentry_streams/adapters/loader.py
fpacifici: You can do this as a follow up

py/sentry_streams/examples/events.py
```diff
@@ -0,0 +1,70 @@
...
+        fake_p95 = max(self.latencies)
+
+        return AlertData(p95_latency=fake_p95, event_count=self.count)
+
+    def merge(self, other: Self) -> Self:
+        # Could use datasketches
+        self.latencies = self.latencies + other.latencies
+        self.count = self.count + other.count
+
+        return self
```
fpacifici, line 70: Add a todo please: this should be managed through a data sketch.

py/sentry_streams/examples/spans.py
```diff
@@ -0,0 +1,83 @@
...
+
+class SpansBuffer(Accumulator[Span, Segment]):
+    """
+    Ingests spans into a window. Builds a Segment from each
+    window, which contains the list of Spans seen as well
+    as the total duration across Spans.
+    """
+
+    def __init__(self) -> None:
+        self.spans_list: list[Span] = []
```
fpacifici, line 66: Spans should be grouped by trace id. A segment is like a transaction. 
Though you can manage this as a todo

py/sentry_streams/window.py
fpacifici, line 1: Same, let's have a pipeline package.

fpacifici (approved):
Please fix the conflicts and address the missing concerns.

PR #20, opened 2025-02-20
add ruff line-length setting for VS Code

bmckerry:
This might be a PR for nobody but me, but this makes VS Code's `ruff` integration work with our line length setting for `black` (so I'm not accidentally reformatting every file I touch)


evanh (approved):

evanh:
Game changer!

PR #21, opened 2025-02-20
feat(flink): create filter step

### LINKED ISSUE #19: Build filter primitive
We have a map primitive
https://github.com/getsentry/streams/blob/main/py/sentry_streams/pipeline.py#L108-L118

Let's build a filter that either send a message through or ignores it.

bmckerry:
This PR adds a filter step to our Flink Adapter.

Closes #19


py/sentry_streams/pipeline.py
```diff
@@ -116,3 +117,15 @@ class Map(WithInput):
...
+
+@dataclass
+class Filter(WithInput):
+    """
+    A simple Filter, taking a single input and either returning it or None as output
+    """
+
+    # TODO: Support a reference to a function (Callable)
+    # instead of a raw string
+    function: str
```
ayirr7, line 130: I've made this change in the Reduce PR, but we can implement this TODO in this PR. https://github.com/getsentry/streams/pull/18/files#diff-5585b877c5c059560f53faffa7afaaf02a1d1ec107b0e968367cb7a4afcb1721R19

We can make both the Map and Filter step take a Callable type (a reference to the actual function) instead of a import path string. 

So the user will just import their actual function in `example_config.py` and call it there

py/sentry_streams/flink/flink_adapter.py
```diff
@@ -28,6 +29,27 @@ def __init__(self, config: MutableMapping[str, Any], env: StreamExecutionEnviron
         self.environment_config = config
         self.env = env
 
+    def assert_return_type(self, function: Callable[[Any], bool], expected_type: type) -> None:
+        return_type = get_annotations(function)["return"]
+        assert (
+            return_type is expected_type
+        ), f"Filter function {function.__name__} must return a bool, got {return_type} instead"
+
+    def get_function_from_step(self, step: Step) -> Any:
```
ayirr7, line 38: Once we use function references instead of import paths in Map / Filter, we shouldn't need this logic around anymore

py/sentry_streams/flink/flink_adapter.py
```diff
@@ -28,6 +29,27 @@ def __init__(self, config: MutableMapping[str, Any], env: StreamExecutionEnviron
         self.environment_config = config
         self.env = env
 
+    def assert_return_type(self, function: Callable[[Any], bool], expected_type: type) -> None:
```
fpacifici, line 32: Not following this type hint. 
`function: Callable[[Any], bool]` means a function that must return a bool. 

If you expect any Callable here to assert, you do not need to provide the bool in the type hint.
Which makes me think: How does mypy actually type check the esages of this function? Does `self.assert_return_type(imported_fn, bool)` actually work ?

py/sentry_streams/flink/flink_adapter.py
```diff
@@ -28,6 +29,27 @@ def __init__(self, config: MutableMapping[str, Any], env: StreamExecutionEnviron
...
+        return_type = get_annotations(function)["return"]
+        assert (
+            return_type is expected_type
+        ), f"Filter function {function.__name__} must return a bool, got {return_type} instead"
+
+    def get_function_from_step(self, step: Step) -> Any:
+        assert hasattr(step, "function")
```
fpacifici, line 39: @ayirr7  I thought we had fixed this with a parent class that have `function` defined. Am I wrong ?

py/sentry_streams/sample_functions.py
fpacifici, line 1: Please let's move all the examples in a package for that.

py/tests/test_pipeline.py
fpacifici, line 1: Please let's have a unit test for the flink adapter. This test does not assert that the adapter produces the right data structure. You may see that the json produced does not mention the filter function.

py/tests/test_pipeline.py
```diff
@@ -191,12 +278,49 @@ def bad_map() -> Pipeline:
...
+        function="sentry_streams.sample_functions.EventsPipelineFunctions.wrong_type_filter",
+    )
+
+    _ = KafkaSink(
+        name="kafkasink",
+        ctx=pipeline,
+        inputs=[map],
+        logical_topic="transformed-events",
+    )
+
```
fpacifici, line 314: You do not need to parametrize the test if there is only one parameter.

fpacifici (changes requested):
Please see comments inline

py/sentry_streams/flink/flink_adapter.py
ayirr7: Separate point, but about the same function, if we decide to keep it around, can we move it to another file? 

For example, I had started a `user_functions` dir here https://github.com/getsentry/streams/pull/18/files#diff-3cd95955acd7db9b632c30d610b47cfc2f436a365faa7ee994db8d14b10f6874. Maybe we can create a similar folder and have function-related utils in another file inside.

py/sentry_streams/flink/flink_adapter.py
ayirr7: That's right, we did add it. @bmckerry can you change the `FlinkAdapter` to match the signature of the `StreamAdapter`? That way, we won't have to perform these assertions since `Map` type has `function` attribute.

ayirr7:
Just checking, did you get a chance to run the pipeline Source --> Filter --> Map --> Sink locally? e.g. using `example_config.py` and see the filtered and transformed data into the output topic?

bmckerry:
> Just checking, did you get a chance to run the pipeline Source --> Filter --> Map --> Sink locally? e.g. using example_config.py and see the filtered and transformed data inside the output Kafka topic?

Yes, I checked locally with `example_config` and the sample filter works as expected

py/sentry_streams/pipeline.py
bmckerry: I've added this for the filter step here: https://github.com/getsentry/streams/pull/21/commits/70c6e7e2603fc7f006ae51d3d222704ac1094efa

py/sentry_streams/flink/flink_adapter.py
bmckerry: I've deleted this function and moved the `assert` into the filter step since passing the function as a callable simplified things

py/sentry_streams/flink/flink_adapter.py
bmckerry: > can you change the FlinkAdapter to match the signature of the StreamAdapter?

Not sure I follow, sorry - they look to have the same signature to me (unless I'm misunderstanding)?
I'll remove the `assert` here since `Map` & `Filter` have `function` as an attribute

ayirr7: Sorry, to be a bit more specific, we want to change `step: Step` to something like `step: Map` for the map() function in the Flink Adapter. Same thing with filter(). LMK if that makes sense

py/sentry_streams/example_config.py
```diff
@@ -9,11 +16,18 @@
     logical_topic="logical-events",
 )
 
+filter = Filter(
+    name="myfilter",
+    ctx=pipeline,
+    inputs=[source],
+    function=EventsPiplineFilterFunctions.simple_filter,
```
ayirr7, line 23: nit on spelling of `Pipeline` :)

py/tests/test_pipeline.py
bmckerry: Discussed offline, we don;t have to worry about this for now

py/sentry_streams/flink/flink_adapter.py
```diff
@@ -83,3 +83,12 @@ def map(self, step: Map, stream: Any) -> Any:
 
         # TODO: Ensure output type is configurable like the schema above
         return stream.map(func=lambda msg: imported_fn(msg), output_type=Types.STRING())
+
+    def filter(self, step: Filter, stream: Any) -> Any:
+        imported_fn = step.function
+        return_type = imported_fn.__annotations__["return"]
+        assert (
+            return_type is bool
+        ), f"Filter function {imported_fn.__name__} must return a bool, got {return_type} instead"
```
fpacifici, line 92: There are two issues here.
1. If you rely on the callable, you do not need to assert the return type. You can Make the type of the function field in the `Filter` class to be a callable that return bool: `Callable[[Any], bool]`. Then the type checker will prevent a function that does not return a boolean to be here.
2. If you support the dynamic function loader (the function is a string) then you have to assert the return type but also load the module dynamically as you were doing before.

py/sentry_streams/pipeline.py
```diff
@@ -124,3 +125,13 @@ class Map(WithInput):
...
     step_type: StepType = StepType.MAP
+
+
+@dataclass
+class Filter(WithInput):
+    """
+    A simple Filter, taking a single input and either returning it or None as output
+    """
+
+    function: Callable[..., Any]
```
fpacifici, line 136: I think this is not the right type. 
1. You still need to support the string version
2. The typed versionshould be a callable that takes a single argument of type Any and returns a bool: `Callable[[Any], bool]` Otherwise the type checker will not complain against any function type provided here.

py/sentry_streams/user_functions/sample_filter.py
```diff
@@ -0,0 +1,17 @@
...
+        return True if "name" in d else False
+
+    @staticmethod
+    def wrong_type_filter(value: str) -> str:
+        # TODO: move test functions into the tests/ folder somehow
+        """
+        Filter with wrong return type, used in tests
+        """
+        d = json.loads(value)
+        return "True" if "name" in d else "False"
```
fpacifici, line 17: If your test is suppsoed to fail here you do not need to provide a body to the function: Just `pass`

py/sentry_streams/flink/flink_adapter.py
bmckerry: I'm opting to only pass the filter function as a callable instead of dynamically loading the function as it seems safer to me

fpacifici: > I'm opting to only pass the filter function as a callable instead of dynamically loading the function as it seems safer to me

If you pass only the Callable, then you have to adjust the `Map` as well as it supports both. We must be consistent.
I'd suggest you support the dynamic loader as well considering there is still debate on whether the Callable should be possible.

py/sentry_streams/user_functions/sample_filter.py
```diff
@@ -0,0 +1,16 @@
+import json
+
+
+class EventsPipelineFilterFunctions:
```
ayirr7, line 4: While we're already adding the filter functions to `user_functions` directory, can you move the `sample_map` functions into the same directory?

Then `user_functions` can be a central place (for now) where users define their custom logic.

py/sentry_streams/flink/flink_adapter.py
```diff
@@ -29,6 +29,30 @@ def __init__(self, config: MutableMapping[str, Any], env: StreamExecutionEnviron
...
+    def load_function(self, step: Step) -> Any:
+        """
+        Takes a transform step containing a function, and either returns
+        the function (if it's a Callable) or dynamically loads and returns the
+        function (if it's a path to a module).
+        """
+        # TODO: break out the dynamic loading logic into a
+        # normalization layer before the flink adapter
+        assert hasattr(step, "function"), "load_function() requires a step containing a function."
+        if isinstance(step.function, str):
```
fpacifici, line 41: Please let's not turn this back to dynamic loading.
`hasattr(step, "function")` 

There are ways to solve this with proper typing:

1. `step: Union[Map, Filter]`. This makes `step.function` statically typed.
2. Create a class on top of Fitler and Map they inherit the `function` attribute from. Then `load_function` can take an object of that type.

fpacifici (approved):

PR #22, opened 2025-02-21
Decouple Flink initialization from the runner

fpacifici:
The current version of the runner is very tightly coupled to Flink.
It initialize Flink in the main method.
This is needed because the pipeline environment is not hidden by the
adapter. 

This makes it impossible to create other adapters and use them in the
same runtime.

This PR moves the Flink StreamingEnvironment management inside the
adapter.


py/sentry_streams/flink/flink_adapter.py
```diff
@@ -25,10 +26,34 @@ class FlinkAdapter(StreamAdapter):
...
+            libs_path = os.environ.get("FLINK_LIBS")
+
+        if libs_path:
+            jar_file = os.path.join(
+                os.path.abspath(libs_path), "flink-sql-connector-kafka-3.4.0-1.20.jar"
+            )
+            kafka_jar_file = os.path.join(
+                os.path.abspath(libs_path), "flink-sql-connector-kafka-3.4.0-1.20.jar"
+            )
+            print(kafka_jar_file)
```
markstory, line 50: Should this be removed?

fpacifici: yes

py/sentry_streams/flink/flink_adapter.py
```diff
@@ -25,10 +26,33 @@ class FlinkAdapter(StreamAdapter):
...
-    def __init__(self, config: MutableMapping[str, Any], env: StreamExecutionEnvironment) -> None:
+    def __init__(self, config: PipelineConfig, env: StreamExecutionEnvironment) -> None:
         self.environment_config = config
         self.env = env
 
+    @classmethod
+    def build(cls, config: PipelineConfig) -> Self:
+        env = StreamExecutionEnvironment.get_execution_environment()
+
+        flink_config = config.get("flink", {})
```
ayirr7, line 37: Is the idea that, at least for now, each streaming pipeline will have a large config dict that stores configuration for each different runtime? 

Do we eventually want to move to defining general configs which can be "translated" into any specific runtime's config?

fpacifici: > Is the idea that, at least for now, each streaming pipeline will have a large config dict that stores configuration for each different runtime?

At this point yes. It may be structured though. 
as we iterate many settings could become generic. So yes to your second question as well.

ayirr7 (approved):

PR #23, opened 2025-02-21
Allow the Map primitive to get a reference to a Callable

fpacifici:
The Map primitive only allows the user to provide the function to
use as a mapper via a fully qualified module name as a string.

While this is convenient and inevitable for pipeline that spans 
different code bases, it is not ergonomic for pipelines where the 
application code and the pipeline definition are in the same
code base.
Here by providing the Map function as a Callable allows for type
checking.

### DIFF
--- a/py/sentry_streams/flink/flink_adapter.py
+++ b/py/sentry_streams/flink/flink_adapter.py
@@ -13,7 +13,7 @@ from pyflink.datastream.connectors.kafka import (
 
 from sentry_streams.adapters.stream_adapter import StreamAdapter
 from sentry_streams.modules import get_module
-from sentry_streams.pipeline import Step
+from sentry_streams.pipeline import Map, Step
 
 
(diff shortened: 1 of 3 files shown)


py/sentry_streams/flink/flink_adapter.py
```diff
@@ -65,20 +65,21 @@ def sink(self, step: Step, stream: Any) -> Any:
 
         return stream.sink_to(sink)
 
-    def map(self, step: Step, stream: Any) -> Any:
+    def map(self, step: Map, stream: Any) -> Any:
+        if isinstance(step.function, str):
```
ayirr7, line 69: Do we still intend to support function paths specified as strings?

ayirr7 (approved):

ayirr7:
I'm assuming you've ran it locally just to check that the pipeline works as expected?

PR #32, opened 2025-02-28
WIP: add broadcast functionality

bmckerry:
This PR adds the ability to broadcast a message from a step to multiple downstream steps

### DIFF
--- a/py/sentry_streams/example_config.py
+++ b/py/sentry_streams/example_config.py
@@ -33,9 +33,30 @@ map = Map(
     function=EventsPipelineMapFunctions.simple_map,
 )
 
-sink = KafkaSink(
-    name="kafkasink",
+branch_1 = Map(
+    name="mybranch1",
     ctx=pipeline,
     inputs=[map],
+    function=EventsPipelineMapFunctions.simple_map,
+)
+
+branch_2 = Map(
+    name="mybranch2",
+    ctx=pipeline,
+    inputs=[map],
+    function=EventsPipelineMapFunctions.simple_map,
+)
+
+sink_1 = KafkaSink(
+    name="kafkasink1",
+    ctx=pipeline,
+    inputs=[branch_1],
     logical_topic="transformed-events",
+)
+
+sink_2 = KafkaSink(
+    name="kafkasink2",
+    ctx=pipeline,
+    inputs=[branch_2],
+    logical_topic="transformed-events-2",
 )
--- a/py/tests/test_pipeline.py
+++ b/py/tests/test_pipeline.py
@@ -18,7 +18,6 @@ from sentry_streams.user_functions.sample_filter import (
 def setup_basic_flink_env() -> (
     Generator[tuple[StreamExecutionEnvironment, RuntimeTranslator], None, None]
 ):
-
     # TODO: read from yaml file
     environment_config = {
         "topics": {
@@ -253,3 +252,56 @@ def test_import(
 
     with pytest.raises(ImportError):
         iterate_edges(pipeline, translator)
+
+
+def broadcast_pipeline() -> Pipeline:
+    pipeline = Pipeline()
+
+    source = KafkaSource(
+        name="myinput",
+        ctx=pipeline,
+        logical_topic="logical-events",
+    )
+
+    map = Map(
+        name="mymap",
+        ctx=pipeline,
+        inputs=[source],
+        function="sentry_streams.unknown_module.EventsPipelineFunctions.simple_map",
+    )
+
+    branch_1 = Map(
+        name="mybranch1",
+        ctx=pipeline,
+        inputs=[map],
+        function=EventsPipelineMapFunctions.simple_map,
+    )
+
+    branch_2 = Map(
+        name="mybranch2",
+        ctx=pipeline,
+        inputs=[map],
+        function=EventsPipelineMapFunctions.simple_map,
+    )
+
+    _ = KafkaSink(
+        name="kafkasink1",
+        ctx=pipeline,
+        inputs=[branch_1],
+        logical_topic="transformed-events",
+    )
+
+    _ = KafkaSink(
+        name="kafkasink2",
+        ctx=pipeline,
+        inputs=[branch_2],
+        logical_topic="transformed-events-2s",
+    )
+
+    return pipeline
+
+
+def test_broadcast(
+    pipeline: Pipeline = broadcast_pipeline(),
+) -> None:
+    assert pipeline.outgoing_edges["mymap"] == ["mybranch1", "mybranch2"]


py/sentry_streams/example_config.py
```diff
@@ -33,9 +34,36 @@
     function=EventsPipelineMapFunctions.simple_map,
 )
 
-sink = KafkaSink(
-    name="kafkasink",
+broadcast = Broadcast(
+    name="broadcast",
     ctx=pipeline,
     inputs=[map],
+)
```
fpacifici, line 41: Why do we need a broadcast step ? 

Isn't this enough 
```
ranch_1 = Map(
    name="mybranch1",
    ctx=pipeline,
    inputs=[map],
    function=EventsPipelineMapFunctions.simple_map,
)

branch_2 = Map(
    name="mybranch2",
    ctx=pipeline,
    inputs=[map],
    function=EventsPipelineMapFunctions.simple_map,
)
```

If two steps have the same input step we should not need an intermediate one.
Other question is whether the absence of an explicit node that performs the broadcast is actually counter intuitive (it may be) but that would be solved by inverting the dependencies rather than adding a node that does nothing semantically. That would be a matter of discussion beyond the scope of this PR.

bmckerry: Having a broadcast step felt more in-line with the rest of the API, although I agree it isn't necessary and am open to implicitly doing broadcast instead. I'll start looking into that.

py/sentry_streams/user_functions/sample_filter.py
```diff
@@ -14,17 +19,3 @@ def wrong_type_filter(value: str) -> str:  # type: ignore[empty-body]
...
-    Sample user-defined functions to
-    plug into pipeline
-    """
-
-    @staticmethod
-    def simple_map(value: str) -> str:
-        d = json.loads(value)
-        res: str = d.get("name", "no name")
-
-        return "hello " + res
```
bmckerry, line 30: this was accidentally included in `sample_filter.py`, I broke map functions out into their own file

py/sentry_streams/runner.py
```diff
@@ -87,13 +79,17 @@ def main() -> None:
...
         },
         "broker": args.broker,
     }
 
     pipeline: Pipeline = pipeline_globals["pipeline"]
     runtime = load_adapter(AdapterType.FLINK, environment_config)
     translator = RuntimeTranslator(runtime)
+    from pprint import pprint
+
+    pprint(pipeline.__dict__)
```
fpacifici, line 92: Please remove

fpacifici (changes requested):
I think it is time to start writing proper unit tests. 
This PR is merged and it introduced a Dummy stream adapter https://github.com/getsentry/streams/commit/3644c772b96dc80e925511987a3cf8409fdd0a9b#diff-e3dbc72f4c52984eb50f72448876045ab92591904f8bb26e9f2d47bf3b2571e4R7-R31.

Which means that can be mocked or new mock adapters can be created for the unit tests.

Now what you can do is to create a unit test where you create a pipeline and call this method passing the dummy adapter. Then you can do one of these two  things:
1. assert the calls to a mock of the dummy adapter are performed.
2. create a dummy adapter (or update the existing one) that keeps a data structure containing the dataflow and write assertions on that.

fpacifici (commented):
I think you should split the sentry_stream part from the sentry_flink part. They are now two separate libraries. 
Please make the change to the sentry_streams part first, you should be able to build that in isolation, then the test will pass.

ayirr7:
Can we add an example pipeline, or, modify an existing pipeline in `examples/` to demonstrate how to use the broadcast functionality? 

You can test with the Dockerized Flink setup to see if the data is ending up in the topic(s)

sentry_flink/tests/test_pipeline.py
```diff
@@ -24,7 +24,6 @@
 def setup_basic_flink_env() -> (
     Generator[tuple[StreamExecutionEnvironment, RuntimeTranslator], None, None]
 ):
-
```
ayirr7, line 27: just a nit to remove extra line

bmckerry: this was a formatting change, ruff didn't like the whitespace there

If you'd prefer I keep it I can find some way to disable it for this repo

ayirr7: it's fine, not a big issue

was just saying that bc it makes the diff cleaner / easier to review but really not that important

sentry_streams/tests/test_runner.py
```diff
@@ -0,0 +1,35 @@
...
+import pytest
+
+from sentry_streams.adapters.stream_adapter import RuntimeTranslator
+from sentry_streams.pipeline import Filter, KafkaSource, Map, Pipeline
+from sentry_streams.runner import iterate_edges
+
+
+@pytest.fixture
+def pipeline() -> Pipeline:
+    pipeline = Pipeline()
```
ayirr7, line 12: A nit but this is a bit difficult to read. Can we change the structure so that it looks more like the Pipeline() initializations we have in `examples/`?

sentry_flink/uv.lock
ayirr7, line 1: we shouldn't need to commit changes to this file IIUC.

sentry_streams/sentry_streams/runner.py
```diff
@@ -36,20 +36,12 @@ def iterate_edges(p_graph: Pipeline, translator: RuntimeTranslator[Stream, Strea
...
-                    output_step_name = output_steps.pop()
-
+                for output in output_steps:
                     # check if the inputs are fanning in
-                    if len(p_graph.incoming_edges[output_step_name]) > 1:
-                        pass
-
-                    # 1:1 between input and output stream
+                    if len(p_graph.incoming_edges[output]) > 1:
+                        print(f"fan-in detected: {output}")
```
fpacifici, line 42: Please remove.

sentry_streams/tests/test_runner.py
```diff
@@ -0,0 +1,35 @@
...
+    translator = RuntimeTranslator(runtime)
+    iterate_edges(pipeline, translator)
+
+    runtime.assert_has_calls(
+        [
+            call.source(pipeline.steps["source1"]),
+            call.map(pipeline.steps["step1"], runtime.source()),
+            call.filter(pipeline.steps["step2"], runtime.map()),
+            call.map(pipeline.steps["step3"], runtime.filter()),
+            call.map(pipeline.steps["step4"], runtime.filter()),
```
fpacifici, line 33: How does this work? 

these calls to `runtime.source()`, `runtime.map()`, etc. are part of the calls on the mock as the expressions are evaluated before `assert_has_calls`

fpacifici (commented):
Please address the missing comments then it is ok.

sentry_streams/tests/test_runner.py
bmckerry: only now do I realize how funny this looks

sentry_streams/tests/test_runner.py
bmckerry: > How does this work?

no idea

All I know is the `assert_has_calls` only passes if I pass those `runtime.*` calls as each`translate_step` call has this `input_stream` argument, which is the previous step: https://github.com/getsentry/streams/blob/main/sentry_streams/sentry_streams/runner.py#L55

bmckerry: I guess `assert_has_calls` is only asserting that a certain ordered subset of the calls exist, not the entire set of calls to the mock object

sentry_streams/sentry_streams/runner.py
```diff
@@ -36,24 +36,12 @@ def iterate_edges(p_graph: Pipeline, translator: RuntimeTranslator[Stream, Strea
...
-                        pass
-
-                    # 1:1 between input and output stream
-                    else:
-                        next_step: WithInput = cast(WithInput, p_graph.steps[output_step_name])
-                        print(f"Apply step: {next_step.name}")
-                        # TODO: Make the typing align with the streams being iterated through. Reconsider algorithm as needed.
-                        next_step_stream = translator.translate_step(next_step, input_stream)  # type: ignore
-                        step_streams[next_step.name] = next_step_stream
+                for output in output_steps:
```
ayirr7, line 39: Thanks for the refactor here

(context) The old structure with if-elses was there just in case we needed to do something special for fan-in or fan-out

sentry_flink/uv.lock
ayirr7, line 1: Let's remove this one too

bmckerry: @ayirr7 I looked at the [docs](https://docs.astral.sh/uv/guides/projects/#uvlock), they make it sound like the uv lockfile should contain the exact versions of packages used by this project and is updated automatically by `uv` - since our project requires setuptools (from our [pyproject](https://github.com/getsentry/streams/blob/main/sentry_flink/pyproject.toml#L2)) I feel like it makes sense to include it in the lockfile? Just so I understand, what's the reasoning for not updating the lockfile?

ayirr7 (approved):
I think you need to merge main and then, if CI is still broken, split up the PR

fpacifici (approved):

PR #51, opened 2025-03-05
fix: post-split javaless flink

bmckerry:
#31 broke the docker-based flink setup because paths in the  dockerfile were hardcoded, this fixes it

### DIFF
--- a/platforms/flink/Dockerfile
+++ b/platforms/flink/Dockerfile
@@ -50,6 +50,6 @@ FROM build_base
 RUN python3.11 -m pip install yq
 ARG PY_PROJECT=sentry_flink/pyproject.toml
 COPY ${PY_PROJECT} ./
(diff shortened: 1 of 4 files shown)


platforms/flink/Dockerfile
```diff
@@ -50,6 +50,6 @@ FROM build_base
 RUN python3.11 -m pip install yq
 ARG PY_PROJECT=sentry_flink/pyproject.toml
 COPY ${PY_PROJECT} ./
-RUN tomlq -r '.project.dependencies[] | select(.path | not)' pyproject.toml | jq -r '.' > ./requirements.txt
+RUN tomlq -r '.project.dependencies' pyproject.toml | jq -r '.[]' > ./requirements.txt
```
bmckerry, line 53: not needed after https://github.com/getsentry/streams/commit/967c8494b09686711990b2f41427d7b59faf7e66

ayirr7 (approved):

PR #53, opened 2025-03-06
ref: split user_functions part 1

bmckerry:
Part 1 of 2 PRs

This copies the `EventsPipelineMapFunctions` from `sample_filter.py` to a new `sample_map.py`.

The old `EventsPipelineMapFunctions` hasn't been deleted and the import paths in `sentry_flink/tests/` haven't been updated because this change will need to be deployed first. Those will be done in part 2.

### DIFF
--- a/sentry_streams/sentry_streams/example_config.py
+++ b/sentry_streams/sentry_streams/example_config.py
@@ -5,10 +5,8 @@ from sentry_streams.pipeline import (
     Map,
     Pipeline,
 )
-from sentry_streams.user_functions.sample_filter import (
-    EventsPipelineFilterFunctions,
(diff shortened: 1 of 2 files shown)


sentry_streams/sentry_streams/example_config.py
```diff
@@ -5,10 +5,8 @@
     Map,
     Pipeline,
 )
-from sentry_streams.user_functions.sample_filter import (
-    EventsPipelineFilterFunctions,
```
victoria-yining-huang, line 9: were you in the `sentry_streams` venv when this formatting happened? I'm pretty sure the venv's black would have formatted back to the original way

victoria-yining-huang (approved):

PR #56, opened 2025-03-07
add more unittests for Pipeline class

victoria-yining-huang:
the existing tests import a pipeline from another file. I remove the dependency and made a new pipeline for these test cases. 

Also added test cases when:
- one step has multiple downstream steps
- multiple steps go to one downstream step

### DIFF
--- a/sentry_streams/tests/test_pipeline.py
+++ b/sentry_streams/tests/test_pipeline.py
@@ -1,5 +1,76 @@
-from sentry_streams.examples.word_counter import pipeline, source
-from sentry_streams.pipeline.pipeline import Step
+import json
+
+from sentry_streams.pipeline.pipeline import (
+    Filter,
+    KafkaSink,
+    KafkaSource,
+    Map,
+    Pipeline,
+    Step,
+)
+
+pipeline = Pipeline()
+
+
+def simple_filter(value: str) -> bool:
+    d = json.loads(value)
+    return True if "name" in d else False
+
+
+def simple_map(value: str) -> str:
+    d = json.loads(value)
+    res: str = d.get("name", "no name")
+
+    return "hello " + res
+
+
+source = KafkaSource(
+    name="myinput",
+    ctx=pipeline,
+    logical_topic="logical-events",
+)
+
+source_2 = KafkaSource(
+    name="myinput2",
+    ctx=pipeline,
+    logical_topic="anotehr-logical-events",
+)
+
+
+filter = Filter(
+    name="myfilter",
+    ctx=pipeline,
+    inputs=[source, source_2],
+    function=simple_filter,
+)
+
+filter_2 = Filter(
+    name="myfilter2",
+    ctx=pipeline,
+    inputs=[filter],
+    function=simple_filter,
+)
+
+map = Map(
+    name="mymap",
+    ctx=pipeline,
+    inputs=[filter],
+    function=simple_map,
+)
+
+map_2 = Map(
+    name="mymap2",
+    ctx=pipeline,
+    inputs=[filter],
+    function=simple_map,
+)
+
+sink = KafkaSink(
+    name="kafkasink",
+    ctx=pipeline,
+    inputs=[map, map_2],
+    logical_topic="transformed-events",
+)
 
 
 def test_register_step() -> None:
@@ -9,8 +80,8 @@ def test_register_step() -> None:
 
 
 def test_register_edge() -> None:
-    assert pipeline.outgoing_edges["myfilter"] == ["mymap"]
-    assert pipeline.incoming_edges["mymap"] == ["myfilter"]
+    assert pipeline.outgoing_edges["myfilter"] == ["myfilter2", "mymap", "mymap2"]
+    assert pipeline.incoming_edges["myfilter"] == ["myinput", "myinput2"]
 
 
 def test_register_source() -> None:
--- a/sentry_streams/tests/test_pipeline.py
+++ b/sentry_streams/tests/test_pipeline.py
@@ -14,40 +14,40 @@ from sentry_streams.pipeline.pipeline import (
 def pipeline() -> Pipeline:
     pipeline = Pipeline()
     source = KafkaSource(
-        name="myinput",
+        name="source",
         ctx=pipeline,
         logical_topic="logical-events",
     )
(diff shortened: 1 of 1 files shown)


sentry_streams/tests/test_pipeline.py
```diff
@@ -9,8 +80,8 @@ def test_register_step() -> None:
 
 
 def test_register_edge() -> None:
-    assert pipeline.outgoing_edges["myfilter"] == ["mymap"]
-    assert pipeline.incoming_edges["mymap"] == ["myfilter"]
+    assert pipeline.outgoing_edges["myfilter"] == ["myfilter2", "mymap", "mymap2"]
```
fpacifici, line 83: Please pass the pipeline with a fixture rather than having a global object.
If you create the pipeline in the fixture it gets re-created at each test so that you can safely make changes to the data structure in the test itself. 
If you share a global object you may make changes in a test and make another unrelated test fail for that. It will be confusing.

sentry_streams/tests/test_pipeline.py
```diff
@@ -1,5 +1,76 @@
...
+    name="myinput",
+    ctx=pipeline,
+    logical_topic="logical-events",
+)
+
+source_2 = KafkaSource(
+    name="myinput2",
+    ctx=pipeline,
+    logical_topic="anotehr-logical-events",
+)
```
fpacifici, line 37: Do you need this in the test? 
You only want to create the objects you need in the test or for the sake of what you are testing so the test is easier to read.
It does not seem anything assert a property on it.

sentry_streams/tests/test_pipeline.py
```diff
@@ -1,5 +1,76 @@
...
+def simple_filter(value: str) -> bool:
+    d = json.loads(value)
+    return True if "name" in d else False
+
+
+def simple_map(value: str) -> str:
+    d = json.loads(value)
+    res: str = d.get("name", "no name")
+
+    return "hello " + res
```
fpacifici, line 24: As you are not actually ever calling these you do need an implementation. That will make the test quicker to read:

```
def simple_filter(value: str) -> bool:
    pass
```

fpacifici (changes requested):
A few small changes

sentry_streams/tests/test_pipeline.py
victoria-yining-huang: addressed

sentry_streams/tests/test_pipeline.py
victoria-yining-huang: addressed

sentry_streams/tests/test_pipeline.py
victoria-yining-huang: made fixture

sentry_streams/tests/test_pipeline.py
```diff
@@ -1,17 +1,91 @@
...
+    KafkaSink(
+        name="kafkasink",
+        ctx=pipeline,
+        inputs=[map, map_2],
+        logical_topic="transformed-events",
+    )
+    return pipeline
+
+
+def simple_filter(value: str) -> bool:
```
ayirr7, line 65: What's the rationale for having these empty tests?

sentry_streams/tests/test_pipeline.py
```diff
@@ -1,17 +1,91 @@
...
+
+    filter = Filter(
+        name="myfilter",
+        ctx=pipeline,
+        inputs=[source, source_2],
+        function=simple_filter,
+    )
+
+    _ = Filter(
+        name="myfilter2",
```
ayirr7, line 36: Kind of a comment on the readability of this but can we rename the names to something simpler? e.g. `map2`. And let's also make sure the variable name matches that hardcoded str name for each step

sentry_streams/tests/test_pipeline.py
```diff
@@ -1,17 +1,91 @@
...
+
+def test_register_step(pipeline: Pipeline) -> None:
     step = Step("new_step", pipeline)
     assert "new_step" in pipeline.steps
     assert pipeline.steps["new_step"] == step
 
 
-def test_register_edge() -> None:
-    assert pipeline.outgoing_edges["myfilter"] == ["mymap"]
+def test_register_edge(pipeline: Pipeline) -> None:
```
ayirr7, line 81: I think there's a couple of other fan-ins and fan-outs to test

map2 <-- [map1, filter1] (incoming)
sink <-- [map1, map2] (incoming)

Can we add those assertions too?

sentry_streams/tests/test_pipeline.py
```diff
@@ -1,17 +1,91 @@
...
+    # when there is only one step going to the next step
     assert pipeline.incoming_edges["mymap"] == ["myfilter"]
+    # when one step goes out to multiple steps
+    assert pipeline.outgoing_edges["myfilter"] == ["myfilter2", "mymap", "mymap2"]
+    # when multiple steps forks into one step
```
ayirr7, line 86: very much a nit, but the phrasing of this comment "forks into" is slightly confusing. Forks intuitively implies a fan-out to me. Can we use something like "fans into"? And "fans out to" for the inverse?

sentry_streams/tests/test_pipeline.py
fpacifici: This is the function used in the pipeline as application logic.
I advised not to add an implementation because the goal of this test is to validate that the translator works. So the application logic is not important.

sentry_streams/tests/test_pipeline.py
victoria-yining-huang: done

sentry_streams/tests/test_pipeline.py
victoria-yining-huang: changed

sentry_streams/tests/test_pipeline.py
victoria-yining-huang: added more tests to make sure there is one outgoing edge and one incoming edge being tested for all three scenarios
- one step goes to another step
- multiple steps fan into one step
- one steps fans out to multiple steps

ayirr7 (approved):

fpacifici (approved):

PR #61, opened 2025-03-11
feat: add router primitive to sentry_streams

bmckerry:
This PR adds the router primitive to the sentry_streams package.

sentry_flink changes will come in another PR


sentry_streams/uv.lock
```diff
@@ -294,7 +295,7 @@ wheels = [
 
 [[package]]
 name = "sentry-streams"
-version = "0.0.1"
+version = "0.0.8"
```
bmckerry, line 298: since `uv.lock` is meant to contain the exact versions of packages that the project uses ([docs](https://docs.astral.sh/uv/guides/projects/#uvlock)), I feel the lockfile with the latest version of `sentry_streams` should be committed

sentry_streams/sentry_streams/runner.py
```diff
@@ -39,9 +39,14 @@ def iterate_edges(p_graph: Pipeline, translator: RuntimeTranslator[Stream, Strea
...
-                    # TODO: Make the typing align with the streams being iterated through. Reconsider algorithm as needed.
                     next_step_stream = translator.translate_step(next_step, input_stream)  # type: ignore
-                    step_streams[next_step.name] = next_step_stream
+                    # if next_step_stream is a map of downstream routes
+                    if isinstance(next_step_stream, MutableMapping):
+                        for stream in next_step_stream:
+                            step_streams[stream] = next_step_stream[stream]
+                    else:
+                        # TODO: Make the typing align with the streams being iterated through. Reconsider algorithm as needed.
+                        step_streams[next_step.name] = next_step_stream
```
bmckerry, line 49: this feels very messy to me as it means `translator.translate_step()` can return either a stream or a map containing streams. not sure of a better way to handle this but very open to ideas

sentry_streams/sentry_streams/pipeline/pipeline.py
```diff
@@ -155,6 +166,42 @@ class Filter(TransformStep[bool]):
...
+    step_type: StepType = StepType.ROUTER
+
+    def __post_init__(self) -> None:
+        super().__post_init__()
+        for branch_name in self.routing_table:
+            # ensure name of branch step matches key of routing table
+            branch_step = self.routing_table[branch_name]
+            assert (
+                branch_name == branch_step.name
+            ), f'Branch name must match its routing table key - please change "{branch_name}" to "{branch_step.name}".'
```
bmckerry, line 201: I'm ensuring the `Branch()` name is the same as its routing table key because if they're different then `translate_step()` won't build the flink graph properly.

I considered just making the routing table be a list of all the `Branch()` objects and just pull `branch.name` anytime I needed a string, but that wouldn't work as we need a way for downstream steps to reference specific branches in the `routing_table` ([example](https://github.com/getsentry/streams/pull/61/files#diff-73a447c195039d8cbd2eb85dfdcb632d17ce0dfb37a5e6d98c1d10f5cbef05b7R45)).

sentry_streams/sentry_streams/examples/routing.py
```diff
@@ -0,0 +1,69 @@
+from sentry_streams.pipeline.pipeline import (
```
bmckerry, line 1: if we want a more 'sentry-specific' example I can try making this into a representation of the replays pipeline (it'd still look mostly like this though, and I think there's value into having the simplest possible version of router since it shows how `Branch` works in a very straightforward way)

sentry_streams/sentry_streams/runner.py
```diff
@@ -39,9 +39,14 @@ def iterate_edges(p_graph: Pipeline, translator: RuntimeTranslator[Stream, Strea
                 for output in output_steps:
                     next_step: WithInput = cast(WithInput, p_graph.steps[output])
                     print(f"Apply step: {next_step.name}")
-                    # TODO: Make the typing align with the streams being iterated through. Reconsider algorithm as needed.
                     next_step_stream = translator.translate_step(next_step, input_stream)  # type: ignore
-                    step_streams[next_step.name] = next_step_stream
+                    # if next_step_stream is a map of downstream routes
+                    if isinstance(next_step_stream, MutableMapping):
+                        for stream in next_step_stream:
```
fpacifici, line 45: rename stream to branch_name. 
Stream seems to indicate the stream object

sentry_streams/sentry_streams/adapters/stream_adapter.py
```diff
@@ -96,6 +98,28 @@ def reduce(
...
+    @abstractmethod
+    def branch(
+        self,
+        step: Branch,
+        stream: Stream,
+    ) -> Stream:
+        """
+        Branch represents a possible branch downstream from a router.
+        """
+        raise NotImplementedError
```
fpacifici, line 121: You do not need the branch form the adapter. This function will not be called. While you need a branch step as a place holder in the `step_streams` object to wire branch steps onto a stream, you do not need to actually translate the Branch step object. That is performed when translating the Router.

sentry_streams/sentry_streams/pipeline/pipeline.py
```diff
@@ -155,6 +166,42 @@ class Filter(TransformStep[bool]):
...
+@dataclass
+class Router(WithInput):
+    """
+    A step which takes a routing table of Branches and sends messages
+    to those branches based on a routing function.
+    Routing functions must only return a single output branch, routing
+    to multiple branches simultaneously is not currently supported.
+    """
+
+    routing_function: Callable[..., str]
```
fpacifici, line 190: Asking to return a string is overly restrictive. People will want to use enums.
Could you please make this generic with respect to the return type of the routing_function. 

It should be possible as the router step or translator does not care in any way about the returned type. It only needs to map the returned type to a Branch.

In the same way `routing_table` should be generic with regards to this type.

sentry_streams/sentry_streams/pipeline/pipeline.py
fpacifici, line 201: Why do we need this ?

I don't think this equality should be needed.
the translator returns a mapping of step name to branch object.
You add the step name with the branch object in the `step_streams` map. The name of the Branch step is not related with the output of the routing function which is the key of the routing table.

sentry_streams/tests/test_pipeline.py
```diff
@@ -91,11 +115,13 @@ def test_register_edge(pipeline: Pipeline) -> None:
...
 def test_register_source(pipeline: Pipeline) -> None:
     assert {pipeline.sources[0].name, pipeline.sources[1].name} == {"source", "source2"}
+
+
+def test_register_branch(pipeline: Pipeline) -> None:
+    assert pipeline.outgoing_edges["router"] == ["branch1", "branch2"]
+    assert pipeline.outgoing_edges["branch1"] == ["kafkasink1"]
+    assert pipeline.outgoing_edges["branch2"] == ["kafkasink2"]
+    assert pipeline.incoming_edges["branch1"] == ["router"]
+    assert pipeline.incoming_edges["branch2"] == ["router"]
```
fpacifici, line 127: You can move these in the same test as above.

sentry_streams/tests/test_runner.py
```diff
@@ -55,5 +72,8 @@ def test_iterate_edges(create_pipeline: Pipeline) -> None:
             call.filter(create_pipeline.steps["step2"], ANY),
             call.map(create_pipeline.steps["step3"], ANY),
             call.map(create_pipeline.steps["step4"], ANY),
+            call.router(create_pipeline.steps["step5"], ANY),
+            call.branch(create_pipeline.steps["step6"], ANY),
+            call.branch(create_pipeline.steps["step7"], ANY),
```
fpacifici, line 77: These should not be called. If they are we have a problem in the way iterate_edges works.

sentry_streams/tests/test_runner.py
```diff
@@ -40,6 +47,16 @@ def create_pipeline() -> Pipeline:
...
     )
+    _ = Router(
+        name="step5",
+        ctx=test_pipeline,
+        inputs=[map],
+        routing_table={
+            "step6": Branch(name="step6", ctx=test_pipeline),
+            "step7": Branch(name="step7", ctx=test_pipeline),
+        },
+        routing_function=lambda x: "branch1",
```
fpacifici, line 58: Please add some following steps so you can verify the methods are called.

sentry_streams/tests/test_runner.py
```diff
@@ -40,6 +47,16 @@ def create_pipeline() -> Pipeline:
...
+    )
```
fpacifici, line 59: The MagicMock adapter is not sufficient anymore for this test as you need to test you are returning the right stream for each branch.
Please modify the DummyAdapter to:
- return something as the stream (the name of the step is enough)
- record all the function calls in a list so that you can run assertions on them
- Use that in the test and assert that the adapter methods are called passing the right stream object and the right parameters.

sentry_streams/sentry_streams/examples/routing.py
fpacifici: Yes we look for realistic tests here.
Please write a pipeline that does this:
1. Starts with a KafkaSource
2. Routes old message to a kafka sink on a BLQ (totally hypothetical)
3. Run a Map 
4. Then add a Broadcast
5. After the broadcast, on one side add a Sink (for super big consumers)
6. On the other branch we would have a Clickhouse sink. For now just a Kafka sink

fpacifici (changes requested):

sentry_streams/sentry_streams/runner.py
```diff
@@ -39,9 +39,14 @@ def iterate_edges(p_graph: Pipeline, translator: RuntimeTranslator[Stream, Strea
...
                     print(f"Apply step: {next_step.name}")
-                    # TODO: Make the typing align with the streams being iterated through. Reconsider algorithm as needed.
                     next_step_stream = translator.translate_step(next_step, input_stream)  # type: ignore
-                    step_streams[next_step.name] = next_step_stream
+                    # if next_step_stream is a map of downstream routes
+                    if isinstance(next_step_stream, MutableMapping):
+                        for branch_name in next_step_stream:
+                            step_streams[branch_name] = next_step_stream[branch_name]
+                    else:
+                        # TODO: Make the typing align with the streams being iterated through. Reconsider algorithm as needed.
```
ayirr7, line 48: nit, This TODO comment had to do with the # type: ignore I left above, can you move it there please

sentry_streams/sentry_streams/adapters/stream_adapter.py
```diff
@@ -96,6 +98,17 @@ def reduce(
         """
         raise NotImplementedError
 
+    @abstractmethod
+    def router(
+        self,
+        step: Router[RoutingFuncReturnType],
+        stream: Stream,
+    ) -> Stream:
```
ayirr7, line 106: Is this typehint accurate now? Router can now also return a map of branch names: branches

sentry_streams/sentry_streams/dummy/dummy_adapter.py
```diff
@@ -10,26 +24,55 @@
...
+        self.add_step_input_streams(step)
         return self
 
+    def reduce(self, step: Reduce[MeasurementUnit, InputType, OutputType], stream: Any) -> Any:
+        self.add_step_input_streams(step)
+        return self
+
+    def router(self, step: Router[RoutingFuncReturnType], stream: Any) -> Any:
+        self.add_step_input_streams(step)
+        return {branch.name: branch for branch in step.routing_table.values()}
```
ayirr7, line 75: Why are we getting the name from the Branch object when the routing table already has the names mapped to the Branches?

bmckerry: I'm going to update my `sentry_flink` router logic to match this - getting the names from the Branch objects means the keys in the router table don't have to be the same as the Branch object names

i.e. we can have a routing_table that looks like
```py
class Branches(Enum):
  NORMAL_BRANCH: "normal_branch"
  WEIRD_BRANCH: "weird_branch"

routing_table = {
  Branches.NORMAL_BRANCH: Branch(name="normal"),
  Branches.WEIRD_BRANCH: Branch(name="weird"),
}
```

In this case if we used the routing table key as the key in the returned map, `iterate_edges()` will fail to get outputs correctly here: https://github.com/getsentry/streams/pull/61/files#diff-e10f98c4977830ef0d37eadbaf2cdf5caa9d4657804ea9f4aaca8e3e5b804859R33

sentry_streams/sentry_streams/dummy/dummy_adapter.py
```diff
@@ -10,26 +24,55 @@
 class DummyAdapter(StreamAdapter[DummyInput, DummyOutput]):
     """
     An infinitely scalable adapter that throws away all the data it gets.
+    The adapter tracks the 'streams' that each step returns in the form of
+    lists of previous step names.
     """
 
     def __init__(self, _: PipelineConfig) -> None:
-        pass
+        self.input_streams: MutableMapping[str, List[str]] = {}
```
ayirr7, line 32: Just trying to make sure I understand the purpose of `input_streams`

Is it going to be used for steps that come after a router and branch? And it basically allows those steps to get the specific stream for the branch / fork that they belong to? 

Like in this example:
Source --> Router --> (Branch1, Branch2) --> (Map1, Map2) 
It allows the Map1 and Map2 to apply their operations on Branch1 and Branch2, respectively?

sentry_streams/sentry_streams/examples/routing.py
fpacifici: Do you plan to build this ?

sentry_streams/sentry_streams/pipeline/pipeline.py
```diff
@@ -155,6 +167,37 @@ class Filter(TransformStep[bool]):
     step_type: StepType = StepType.FILTER
 
 
+@dataclass
+class Branch(Step):
+    """
+    A Branch represents one branch in a pipeline, which is routed to
+    by a Router.
+    The name of the Branch step must match the key of the step in
+    its Router's routing table.
```
fpacifici, line 176: I don't think this is valid anymore

sentry_streams/sentry_streams/pipeline/pipeline.py
```diff
@@ -155,6 +167,37 @@ class Filter(TransformStep[bool]):
...
+class Router(WithInput, Generic[RoutingFuncReturnType]):
+    """
+    A step which takes a routing table of Branches and sends messages
+    to those branches based on a routing function.
+    Routing functions must only return a single output branch, routing
+    to multiple branches simultaneously is not currently supported.
+    """
+
+    routing_function: Callable[..., RoutingFuncReturnType]
+    routing_table: Mapping[str, Branch]
```
fpacifici, line 192: Isn't this supposed to be 
```
routing_table: Mapping[RoutingFuncReturnType, Branch]
```

sentry_streams/sentry_streams/runner.py
```diff
@@ -39,9 +39,14 @@ def iterate_edges(p_graph: Pipeline, translator: RuntimeTranslator[Stream, Strea
                 for output in output_steps:
                     next_step: WithInput = cast(WithInput, p_graph.steps[output])
                     print(f"Apply step: {next_step.name}")
-                    # TODO: Make the typing align with the streams being iterated through. Reconsider algorithm as needed.
                     next_step_stream = translator.translate_step(next_step, input_stream)  # type: ignore
-                    step_streams[next_step.name] = next_step_stream
+                    # if next_step_stream is a map of downstream routes
+                    if isinstance(next_step_stream, MutableMapping):
```
fpacifici, line 44:
```
if isinstance(next_step_stream, Mapping):
```

You do not care about the mapping being mutable. MutableMapping extends Mapping

sentry_streams/sentry_streams/dummy/dummy_adapter.py
```diff
@@ -10,26 +24,55 @@
...
+
+    def add_step_input_streams(self, step: Any) -> None:
```
fpacifici, line 34: Please give this method a more self explanatory name.
What does it do ?
`add_step_input_streams` 

Also step is probably not Any.

sentry_streams/sentry_streams/dummy/dummy_adapter.py
```diff
@@ -10,26 +24,55 @@
...
+        # TODO: update to support multiple inputs to a step
+        # once we implement Union
```
fpacifici, line 36: Please assert that len(inputs) == 1 so we do not fail silently

sentry_streams/sentry_streams/dummy/dummy_adapter.py
```diff
@@ -10,26 +24,55 @@
...
+        input_step = step.inputs[0]
+        input_step_name = input_step.name
+        input_step_stream = self.input_streams[input_step_name]
+        self.input_streams[step.name] = input_step_stream + [input_step_name]
+        # if step is a Router, also add its Branch nodes to input_streams
+        if step.step_type == StepType.ROUTER:
+            for branch in step.routing_table.values():
+                self.input_streams[branch.name] = self.input_streams[step.name] + [step.name]
+        print(f"{step.name=}")
+        print(self.input_streams, "\n")
```
fpacifici, line 47: please remove

sentry_streams/sentry_streams/dummy/dummy_adapter.py
fpacifici: Still not following. 
Why do you need to use the name of the Branch object at all ?

sentry_streams/sentry_streams/examples/routing.py
bmckerry: yes, working on this now

sentry_streams/sentry_streams/dummy/dummy_adapter.py
bmckerry: yes, it can be used to ensure steps downstream from a split in the stream are being applied to the correct stream.
So in your example
```
Source --> Router --> (Branch1, Branch2) --> (Map1, Map2)
```
if we have `Branch1 --> Map1` and `Branch2 --> Map2`, we can verify that 
```
input_streams["Map1"] == [Source, Router, Branch1]
input_streams["Map2"] == [Source, Router, Branch2]
```
I'll add a docstring with some explanation.

ayirr7: Makes sense. Indeed, a docstring will be quite helpful

sentry_streams/sentry_streams/examples/blq_fn.py
```diff
@@ -0,0 +1,35 @@
...
+
+@dataclass
+class KafkaMessage:
+    key: str
+    headers: dict[str, str]
+    value: str
+    timestamp: datetime
+
+    @classmethod
+    def unpack(cls, msg: str) -> "KafkaMessage":
```
fpacifici, line 20: Please add `from __future__ import annotations` then you can use the name of a class declared later.
`def unpack(cls, msg: str) -> KafkaMessage:`

sentry_streams/sentry_streams/examples/blq_fn.py
```diff
@@ -0,0 +1,35 @@
...
+
+class DownstreamBranch(Enum):
+    DELAYED = "delayed"
+    RECENT = "recent"
+
+
+@dataclass
+class KafkaMessage:
+    key: str
+    headers: dict[str, str]
```
fpacifici, line 15: I don't think we 'd want to propagate these Kafka implementation details through the pipeline. Also we will not want to call the message class KafkaMessage.

Since this is an example we use it to showcase the api. 
Please rename it to Message and remove key and headers.

sentry_streams/sentry_streams/adapters/stream_adapter.py
```diff
@@ -111,7 +111,7 @@ def router(
         self,
         step: Router[RoutingFuncReturnType],
         stream: Stream,
-    ) -> Mapping[RoutingFuncReturnType, Stream]:
+    ) -> Mapping[str, Stream]:
```
bmckerry, line 114: I changed this back to `Mapping[str, Stream]` as `router()` returns the branch name as the key of the map, not the routing table key. This is because `iterate_edges()` requires the step name as the key in the `step_streams` map to work properly.

sentry_streams/sentry_streams/adapters/stream_adapter.py
```diff
@@ -104,6 +106,17 @@ def reduce(
         """
         raise NotImplementedError
 
+    @abstractmethod
+    def router(
+        self,
+        step: Router[RoutingFuncReturnType],
+        stream: Stream,
+    ) -> Mapping[str, Stream]:
```
bmckerry, line 114: I changed this back to `Mapping[str, Stream]` as router() returns the branch name as the key of the map, not the routing table key. This is because `iterate_edges()` requires the step name as the key in the `step_streams` map to work properly.

sentry_streams/sentry_streams/dummy/dummy_adapter.py
bmckerry: Discussed with Filippo offline, I'm going to keep using `branch.name` as the key of the returned map because `iterate_edges()` expects that to be the key of the entry in the `step_streams` map here: https://github.com/getsentry/streams/blob/main/sentry_streams/sentry_streams/runner.py#L33

sentry_streams/sentry_streams/dummy/dummy_adapter.py
```diff
@@ -10,26 +25,74 @@
...
+        ), "Only steps with a single input are supported for DummyAdapter."
+
+        input_step = step.inputs[0]
+        input_step_name = input_step.name
+        input_step_stream = self.input_streams[input_step_name]
+        self.input_streams[step.name] = input_step_stream + [input_step_name]
+        # if step is a Router, also add its Branch nodes to input_streams
+        if isinstance(step, Router):
+            for branch in step.routing_table.values():
+                self.input_streams[branch.name] = self.input_streams[step.name] + [step.name]
```
fpacifici, line 63: Please avoid this isinstance. You do not need that.

Jsut put the Router specific logic into the `router` method after calling `track_input_stream` where it belongs.
If you already have multiple methods, each of which contain the responsibility to manage a step, but then always call the same general method that split the responsibilities again, the abstraction is working against you rather than for you.

fpacifici (approved):

PR #77, opened 2025-03-19
feat: Add deployment config

ayirr7:
Adds deployment configuration files which operate on a per-pipeline basis. Each file has an `env` section for general configuration (e.g. topics), a `pipeline` section for the local / dev configs, and allows for additional runtime-specific sections. For example, there is a `flink-override` section if there's something very specific to flink that needs to be configured. 

The Arroyo adapter only uses the config to set up source and sink at the moment. The Arroyo adapter deals with one segment at a time. A segment must be passed into the runner when using the Arroyo adapter. 

The Flink Adapter can now use the config to set up chaining, rebalance in between chains, and set parallelism for a chain. It can take in a whole pipeline with multiple segments. Works with a single worker node setup. From what I understand, there's no way to specifically tell Flink to deploy operators across workers. You can only use `chaining` to enforce them to be on the same worker with no cross-task serialization/deserialization.


sentry_streams/sentry_streams/deployment_config/alerts_config.yaml
```diff
@@ -0,0 +1,37 @@
...
+
+  sinks_config:
+    kafkasink:
+      bootstrap_servers: localhost:9092
+
+  reduce_config:
+    myreduce:
+      parallelism: 3
+
+flink:
```
ayirr7, line 23: This is meant to be for runtime-specific overrides. I'd like to override the fields that are defined in this section -- wanted to use Pydantic for some utils that make operations like deep updates of dictionaries easier, but I can't manage to install it.

ayirr7:
Haven't split the PR yet because I think seeing whole diff is useful

sentry_streams/sentry_streams/deployment_config/alerts_config.yaml
fpacifici, line 1: I think we are missing the concept of pipeline segments to define where we break chains.
Plus I am not sure the adapter specific config should be allowed to override any parameter. For example the flink version should not be able to override the kafka broker config. This makes me think that should be a separation between adapter specific config vs generic config.

What about something like this:

```mermaid
  graph TD;
     subgraph pipeline
       subgraph segment
          subgraph step
              common_config
              subgraph adapter_config
                  adapter1
                  adapter2
              end
           end
       end
    end
```

Where a pipeline is composed of semgnets (or chains) chains are wired up together in the same worker.
In each chain we provide config for each step (that needs config)
each step config can have a common config and a list of config elements per adapter.

There should be a way to provide some common environment like you did to map streams to topics.

sentry_streams/sentry_streams/deployment_config/alerts_config.yaml
```diff
@@ -0,0 +1,37 @@
...
+  parallelism: 2
+  sources_config:
+    myinput:
```
fpacifici, line 26: In https://github.com/getsentry/streams/pull/79 I abstracted away the type of source/sink. So we should pick the right source type in configuration. So you should have a representation of KafkaSource (with kafka paramters) and opening up the possibility to have other types.

sentry_streams/sentry_streams/deployment_config/alerts_config.yaml
```diff
@@ -0,0 +1,37 @@
...
+      consumer_group: python-test
+      auto_offset_reset: latest
+
+  sinks_config:
+    kafkasink:
+      bootstrap_servers: kafka:9093
+
+  reduce_config:
+    myreduce:
+      parallelism: 3
```
fpacifici, line 37: Isn't parallelism a property of a segment that contains a sequence of steps.

fpacifici (commented):
A couple of high level suggestions:
- Define a schema for the config so we can validate it upfront rather than fail when a missing field is loaded. You can use jsonschema (yes. it is yaml but json is a subset so you can use jsonschema to validate).
- Let's have a config class that provides some solid abstracitons to the specific conmponents (i's recommend TypedDict that are lightweight but still type checkable).

platforms/flink/docker-compose.yml
```diff
@@ -10,7 +10,7 @@ x-flink-config: &flink-config |
   jobmanager.memory.process.size: 1024m
   jobmanager.rpc.address: jobmanager
   taskmanager.memory.process.size: 1024m
-  taskmanager.numberOfTaskSlots: 2
+  taskmanager.numberOfTaskSlots: 10
```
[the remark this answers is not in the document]
ayirr7: I did this so I can test out parallelism of steps with Flink

ayirr7:
description was updated (please read)

sentry_streams/sentry_streams/deployment_config/transformer_config.yaml
```diff
@@ -0,0 +1,37 @@
...
+          common:
+            type: source
+            bootstrap_servers: localhost:9092
+            consumer_group: python-test
+            auto_offset_reset: latest
+          flink_overrides:
+
+        parser:
+          common:
+            type: map
```
ayirr7, line 23: These `type` fields are only needed for `source` and `sink` steps. This is because the way the Arroyo adapter works, where we set up the Kafka consumer and producer configuration at the `build` step, instead of doing this when we are processing a Source or Sink Step.

fpacifici:
> From what I understand, there's no way to specifically tell Flink to deploy operators across workers. You can only use chaining to enforce them to be on the same worker with no cross-task serialization/deserialization.

Which of these scenarios are you referring to with "there's no way to specifically tell Flink to deploy operators across workers" ?
- split a chain of two operators over 2 slots preserving cardinality (imagine 2 consumer slot + 2 processing slot for a map operator)
- split a chain of two operators over n slots on the same worker (imagine 2 consumers reshuffling data towards 16 processing slots all running on the same node) 
- split a chain of two operators over n slots across many workers (imagine 2 consumers reshuffling data towards 16 processing slots across 8 workers).
- others?

The ones I provided above are achievable in flink.

fpacifici (changes requested):
Overall a few high level ideas then please see comments inline:
- Let's focus on sensible defaults. If a user is making changes to their pipeline and there is nothing to customize they should not touch the config file. Only touch the config file if there is some information to add by doing so.

sentry_streams/sentry_streams/adapters/arroyo/adapter.py
```diff
@@ -36,6 +36,13 @@
 from sentry_streams.pipeline.window import MeasurementUnit
 
 
+class StepConfig(TypedDict):
+    starts_segment: Optional[bool]
+    ends_segment: Optional[bool]
```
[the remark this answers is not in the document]
ayirr7: Sorry that's old code. It's not actually required if you see the config file examples

sentry_streams/sentry_streams/adapters/arroyo/adapter.py
```diff
@@ -36,6 +36,13 @@
...
+    common: Mapping[str, Any]
+    flink_overrides: Mapping[str, Any]
```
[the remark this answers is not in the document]
ayirr7: I've decided to keep the adapters in separate files. There is a base file which is for the local / dev adapter. 

Right now, in the configs, there is nothing structurally different from Flink config vs default local adapter config. 

However, when that changes, we will introduce a new schema that the Flink ADapter can use to resolve its specific config. Wdyt?

sentry_streams/sentry_streams/adapters/arroyo/adapter.py
```diff
@@ -74,12 +86,13 @@ def add_source(self, step: Source) -> None:
...
         assert isinstance(step, StreamSource), "Only Stream Sources are supported"
         source_name = step.name
+
         if source_name not in self.__sources:
             config = self.__sources_config.get(source_name)
             assert config, f"Config not provided for source {source_name}"
             self.__sources[source_name] = KafkaConsumer(
                 build_kafka_consumer_configuration(
-                    default_config=config["additional_settings"],
+                    default_config={},
```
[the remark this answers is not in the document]
ayirr7: will change

sentry_streams/sentry_streams/adapters/arroyo/adapter.py
```diff
@@ -74,15 +66,21 @@ def add_source(self, step: Source) -> None:
...
+
         if source_name not in self.__sources:
-            config = self.__sources_config.get(source_name)
-            assert config, f"Config not provided for source {source_name}"
+
+            source_config = self.config.get(source_name)
+            assert source_config is not None, f"Config not provided for source {source_name}"
+
+            source_config = cast(KafkaConsumerConfig, source_config)
+            print(source_config)
```
fpacifici, line 76: please remove

sentry_streams/sentry_streams/adapters/loader.py
```diff
@@ -28,6 +30,9 @@ def load_adapter(adapter_type: str, config: PipelineConfig) -> StreamAdapter[Str
     #TODO: Actually move out Flink otherwise everything stated above makes
     # no sense.
     """
+    if segment_id is not None:
+        config = {"env": config["env"], **config["pipeline"]["segments"][segment_id]}
+
```
fpacifici, line 35: is segment id a relevant field for an adapter like Flink ? 
It is alright here. Though food for thoughts

sentry_streams/sentry_streams/deployment_config/flink_transformer_config.yaml
```diff
@@ -0,0 +1,36 @@
...
+      # start the segment
+      {
+      parallelism: 3,
+      steps_config: {
+          kafkasink: {
+          bootstrap_servers: "kafka:9093",
+        }
+      }
+      }
+    ]
```
fpacifici, line 36: I think there is a better way to format this:

```
env:
  topics:
    events: events
    transformed-events: transformed-events
    transformed-events-2: transformed-events-2
pipeline:
  segments:
    - parallelism: 2
      steps_config:
        myinput:
          starts_segment: true
          bootstrap_servers: 'kafka:9093'
    - parallelism: 3
      steps_config:
        kafkasink:
          bootstrap_servers: 'kafka:9093'
```

fpacifici (approved):

PR #79, opened 2025-03-19
Make streams less Kafkaesque

fpacifici:
Our goal is to provide an abstraction over streams. 
We are hiding all the aspects of Kafka like message, commit policy, partitions etc.
Why having to call a stream source KafkaSource when the user only needs to care of a sterm of events ?

### DIFF
--- a/sentry_streams/sentry_streams/adapters/arroyo/adapter.py
+++ b/sentry_streams/sentry_streams/adapters/arroyo/adapter.py
@@ -86,7 +86,7 @@ class StreamSources:
                 )
             )
 
-        self.__source_topics[source_name] = Topic(step.stream)
+        self.__source_topics[source_name] = Topic(step.stream_name)
 
     def get_topic(self, source: str) -> Topic:
         return self.__source_topics[source]
@@ -167,7 +167,7 @@ class ArroyoAdapter(StreamAdapter[Route, Route]):
         ), f"Stream starting at source {stream.source} not found when adding a producer"
 
         self.__consumers[stream.source].add_step(
-            StreamSinkStep(route=stream, producer=producer, topic_name=step.stream)
+            StreamSinkStep(route=stream, producer=producer, topic_name=step.stream_name)
         )
 
         return stream
--- a/sentry_streams/sentry_streams/adapters/arroyo/adapter.py
+++ b/sentry_streams/sentry_streams/adapters/arroyo/adapter.py
@@ -72,7 +72,7 @@ class StreamSources:
         """
         # TODO: Provide a better way to get the logical stream name from
         # the Sink step. We should not have to assert it is a Kafka sink
-        assert isinstance(step, StreamSource), "Only Kafka Sources are supported"
+        assert isinstance(step, StreamSource), "Only Stream Sources are supported"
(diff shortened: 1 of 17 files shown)


sentry_streams/sentry_streams/adapters/arroyo/adapter.py
```diff
@@ -146,7 +146,7 @@ def sink(self, step: Sink, stream: Route) -> Route:
         """
         # TODO: Provide a better way to get the logical stream name from
         # the Sink step. We should not have to assert it is a Kafka sink
-        assert isinstance(step, KafkaSink), "Only Kafka Sinks are supported"
+        assert isinstance(step, StreamSink), "Only Kafka Sinks are supported"
```
bmckerry, line 149: should we change the error msg here?

bmckerry (approved):

sentry_streams/sentry_streams/pipeline/pipeline.py
```diff
@@ -84,12 +84,12 @@ class Source(Step):
 
 
 @dataclass
-class KafkaSource(Source):
+class StreamSource(Source):
     """
     A Source which reads from Kafka.
```
victoria-yining-huang, line 89: i'm not understanding why this entire PR is needed. It seems like this class would still represent a Kafka source? Then why change the name? I understand the streaming platform need to support multiple sources but a KafkaSource is still needed?

fpacifici: If we manage to abstract the source. For now that seems the case, binding the stream source to a specific source type can go in the configuration. 
The fewer technology specific detail we expose the better. The fewer details the product engineers has to think about. In practice, what does the engineer get from knowing this is a Kafka topic ?

It may be that, later on we will figure out that nuance is needed and we will add it back. At this point it does not provide any value.

sentry_streams/sentry_streams/examples/alerts.py
```diff
@@ -61,9 +61,9 @@
     function=build_alert_json,
 )
 
-sink = KafkaSink(
+sink = StreamSink(
     name="kafkasink",
     ctx=pipeline,
     inputs=[map_str],
-    logical_topic="transformed-events",
+    stream="transformed-events",
```
victoria-yining-huang, line 68: `stream` is way too overused in this project and I would want to change this. I don't have a suggestion tho because I also can't think of anything better

fpacifici: Why is it overused? This primitives is referencing a specific stream among all. It seems an appropriate naming. 
Would it be better if it was called `stream_name` ?

victoria-yining-huang: `stream_name` could work too but just to confirm this is just for human readability right?

fpacifici: `stream_slug` that is not a slug. 
YEs, it is jsut for human readability

victoria-yining-huang: sure `steam_name` works!

PR #82, opened 2025-03-20
Introduce chained primitives to replace the old DSL

fpacifici:
The team discussed multiple options on how to represent pipelines in the python DSL.
The winning option was to use a Flink like chaining syntax that has a few advantages:
- It makes the topology evident as each step is chained to the previous one as a
  function call on the return value of the previous one.
- It makes routing and broadcasting much easier.

This PR introduces the DSL and supports all the existing primitives.
Please see the example to see how the syntax looks like.

The transition will be done in three stages:
- This one introduces the new DSL as a way to produce the existing pipeline data structure.
  The chained DSL would introduce fundamental changes to the translator. It becomes
  easier. But I cannto change that as well in this PR
- Port all the examples. I ported two here.
- Replace the translator with one designed for this DSL and update the adapters.

Functionally, what is still missing (will do in a separate PR) is supporting
multi-source pipelines. Now multiple pipeline objects would be created.

### DIFF
--- a/sentry_streams/sentry_streams/examples/blq.py
+++ b/sentry_streams/sentry_streams/examples/blq.py
@@ -30,7 +30,7 @@ save_delayed_message = (
 pipeline = (
     streaming_source(
         name="ingest",
-        stream_name="logical-events",
+        stream_name="events",
     )
     .apply("unpack_message", Map(unpack_kafka_message))
     .route(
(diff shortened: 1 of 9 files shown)


sentry_streams/sentry_streams/examples/blq.py
```diff
@@ -4,73 +4,43 @@
...
+    .sink(
+        "delayed_msg_sinkStreamSource",
+        stream_name="transformed-eventStreamSourceStreamSource",
+    )
+)
+
+pipeline = (
+    streaming_source(
+        name="ingest",
+        stream_name="logical-events",
```
bmckerry, line 35: you'll need to rename all instances of `logical-events` to `events` after https://github.com/getsentry/streams/pull/84
```suggestion
        stream_name="events",
```

bmckerry (approved):

PR #83, opened 2025-03-24
feat: router for arroyo adapter

bmckerry:
This PR adds a router step to the arroyo adapter.

Since arroyo has no concept of branches, adding a router is a bit janky:
- each message payload is wrapped in a `RoutedValue` class, which tracks the `source` of the message and also a list of `waypoints` (representing the list of branch decisions by router steps the message passed through)
- the Router step updates the `waypoints` list of a message based on the outcome of a routing function
- if a message's `Route` matches that of a step's `Route`, the message is processed by the step, otherwise it skips processing and goes on to the next step

Since each branch ends in a sink, this PR introduces the `Forwarder` custom strategy, which will either pass a message to a `Produce` step (if the `Route` of the sink step is correct), or forward the message to the next step in the pipeline (which in our case should be another `SinkStep` since all possible sinks should be the last steps in a pipeline).

### DIFF
--- a/sentry_streams/sentry_streams/adapters/arroyo/adapter.py
+++ b/sentry_streams/sentry_streams/adapters/arroyo/adapter.py
@@ -1,6 +1,15 @@
 from __future__ import annotations
 
-from typing import Any, Mapping, MutableMapping, Self, TypedDict
+from typing import (
+    Any,
+    List,
+    Mapping,
+    MutableMapping,
+    MutableSequence,
+    Self,
+    TypedDict,
+    cast,
+)
 
 from arroyo.backends.kafka.configuration import (
     build_kafka_configuration,
@@ -229,12 +238,15 @@ class ArroyoAdapter(StreamAdapter[Route, Route]):
         """
         assert (
             stream.source in self.__consumers
-        ), f"Stream starting at source {stream.source} not found when adding a filter"
+        ), f"Stream starting at source {stream.source} not found when adding a router"
         self.__consumers[stream.source].add_step(RouterStep(route=stream, pipeline_step=step))
 
         routes_map: MutableMapping[str, Route] = {}
         for branch in step.routing_table.values():
-            branch_stream = Route(source=stream.source, waypoints=stream.waypoints + [branch.name])
+            branch_waypoints = cast(List[str], stream.waypoints) + [branch.name]
+            branch_stream = Route(
+                source=stream.source, waypoints=cast(MutableSequence[str], branch_waypoints)
+            )
             routes_map[branch.name] = branch_stream
 
         return routes_map
--- a/sentry_streams/sentry_streams/adapters/arroyo/adapter.py
+++ b/sentry_streams/sentry_streams/adapters/arroyo/adapter.py
@@ -57,11 +57,13 @@ class StreamSources:
     def __init__(
         self,
         sources_config: Mapping[str, KafkaConsumerConfig],
+        topics_config: Mapping[str, str],
         sources_override: Mapping[str, KafkaConsumer] = {},
     ) -> None:
         super().__init__()
 
         self.__sources_config = sources_config
+        self.__topics_config = topics_config
 
         # Overrides are for unit testing purposes
         self.__source_topics: MutableMapping[str, Topic] = {}
@@ -91,7 +93,7 @@ class StreamSources:
                 )
             )
 
-        self.__source_topics[source_name] = Topic(step.stream_name)
+        self.__source_topics[source_name] = Topic(self.__topics_config[step.stream_name])
 
     def get_topic(self, source: str) -> Topic:
         return self.__source_topics[source]
@@ -105,13 +107,14 @@ class ArroyoAdapter(StreamAdapter[Route, Route]):
         self,
         sources_config: Mapping[str, KafkaConsumerConfig],
         sinks_config: Mapping[str, KafkaProducerConfig],
+        topics_config: Mapping[str, str],
         sources_override: Mapping[str, KafkaConsumer] = {},
         sinks_override: Mapping[str, KafkaProducer] = {},
     ) -> None:
         super().__init__()
-
-        self.__sources = StreamSources(sources_config, sources_override)
+        self.__sources = StreamSources(sources_config, topics_config, sources_override)
         self.__sinks_config = sinks_config
+        self.topics_config = topics_config
 
         # Overrides are for unit testing purposes
         self.__sinks: MutableMapping[str, Any] = {**sinks_override}
@@ -124,6 +127,7 @@ class ArroyoAdapter(StreamAdapter[Route, Route]):
         return cls(
             config["sources_config"],
             config["sinks_config"],
+            config["topics"],
             config.get("sources_override", {}),
             config.get("sinks_override", {}),
         )
@@ -227,7 +231,13 @@ class ArroyoAdapter(StreamAdapter[Route, Route]):
             stream.source in self.__consumers
         ), f"Stream starting at source {stream.source} not found when adding a filter"
         self.__consumers[stream.source].add_step(RouterStep(route=stream, pipeline_step=step))
-        raise NotImplementedError
+
+        routes_map: MutableMapping[str, Route] = {}
+        for branch in step.routing_table.values():
+            branch_stream = Route(source=stream.source, waypoints=stream.waypoints + [branch.name])
+            routes_map[branch.name] = branch_stream
+
+        return routes_map
 
     def get_processor(self, source: str) -> StreamProcessor[KafkaPayload]:
         """
--- a/sentry_streams/sentry_streams/adapters/arroyo/adapter.py
+++ b/sentry_streams/sentry_streams/adapters/arroyo/adapter.py
@@ -66,13 +66,11 @@ class StreamSources:
     def __init__(
         self,
         sources_config: Mapping[str, KafkaConsumerConfig],
-        topics_config: Mapping[str, str],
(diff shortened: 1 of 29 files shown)


sentry_streams/sentry_streams/adapters/arroyo/custom_strategies.py
fpacifici, line 1: nit: please rename this module to something that is specific to the strategy it provides: like call it forwarder.py

Generic bags of unspecified concepts are not great module/package/class names (look like utils classes) as they do not convey the role they have in the architecture.

sentry_streams/sentry_streams/adapters/arroyo/routes.py
```diff
@@ -17,7 +17,7 @@ class Route:
     """
 
     source: str
-    waypoints: Sequence[str]
+    waypoints: List[str]
```
fpacifici, line 20: MutableSequence if you need to mutate it.

sentry_streams/sentry_streams/adapters/arroyo/custom_strategies.py
```diff
@@ -0,0 +1,51 @@
...
+class Forwarder(ProcessingStrategy[Any]):
+    """
+    Either produces an incoming message via a given Producer if the Route of the message
+    matches this strategy's Route, or forwards the message to the next strategy provided.
+    """
+
+    def __init__(
+        self,
+        route: Route,
+        produce_step: Produce[Any],
```
fpacifici, line 20: This should work with `Produce[KafkaPayload]`. Does it ?

sentry_streams/sentry_streams/adapters/arroyo/custom_strategies.py
```diff
@@ -0,0 +1,51 @@
...
+
+from arroyo.backends.kafka.consumer import KafkaPayload
+from arroyo.processing.strategies import Produce
+from arroyo.processing.strategies.abstract import ProcessingStrategy
+from arroyo.types import FilteredPayload, Message
+
+from sentry_streams.adapters.arroyo.routes import Route, RoutedValue
+
+
+class Forwarder(ProcessingStrategy[Any]):
```
fpacifici, line 11: Does this work as `ProcessingStrategy[Union[FilteredPayload, RoutedValue]]`  ?

sentry_streams/sentry_streams/adapters/arroyo/custom_strategies.py
```diff
@@ -0,0 +1,51 @@
...
+        next_step: ProcessingStrategy[Union[FilteredPayload, RoutedValue]],
+    ) -> None:
+        self.__produce_step = produce_step
+        self.__next_step = next_step
+        self.__route = route
+
+    def submit(self, message: Message[Union[FilteredPayload, RoutedValue]]) -> None:
+        message_payload = message.value.payload
+        if isinstance(message_payload, RoutedValue) and message_payload.route == self.__route:
+            kafka_payload = message.value.replace(
```
fpacifici, line 30: Why replacing rather than creating a new Message object ?

sentry_streams/sentry_streams/adapters/arroyo/steps.py
```diff
@@ -126,28 +124,63 @@ def transformer(
...
-    It filters out messages for the route not specified on this step and unpacks
-    the routed message into the original Arroyo payload.
+    Since Arroyo has no concept of 'branches', this updates the `waypoints` list within
+    a message's `Route` object based on the result of the routing function.
     """
 
-    producer: Producer[Any]
-    topic_name: str
+    # TODO: fix generic typing here
+    pipeline_step: Router[RoutingFuncReturnType]  # type: ignore
```
fpacifici, line 137: What's the type issue here ?

sentry_streams/sentry_streams/adapters/arroyo/steps.py
```diff
@@ -126,28 +124,63 @@ def transformer(
...
+            return payload
+
+        def transformer(
+            message: Message[Union[FilteredPayload, RoutedValue]],
+        ) -> Union[FilteredPayload, RoutedValue]:
+            return process_message(
+                self.route,
+                message,
+                append_branch_to_waypoints,
+            )
```
fpacifici, line 159: Nit: this may be more concise if you inline it in a lambda in the call to RunTask:
```
return RunTask(
            lambda msg: process_message(
                self.route,
                message,
                append_branch_to_waypoints,
            ),
            next,
        )
```

sentry_streams/sentry_streams/adapters/arroyo/adapter.py
```diff
@@ -86,7 +93,7 @@ def add_source(self, step: Source) -> None:
                 )
             )
 
-        self.__source_topics[source_name] = Topic(step.stream_name)
+        self.__source_topics[source_name] = Topic(self.__topics_config[step.stream_name])
```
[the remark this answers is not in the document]
fpacifici: Not sure I follow the reasoning here. 
Do we need to introduce the override to make the router work ?
If not please revert this. It is a different topic that still needs to be discussed, you do not want it to be on the critical path to your PR.

sentry_streams/sentry_streams/adapters/arroyo/adapter.py
```diff
@@ -218,7 +227,17 @@ def router(
         """
         Build a router operator for the platform the adapter supports.
         """
-        raise NotImplementedError
+        assert (
+            stream.source in self.__consumers
+        ), f"Stream starting at source {stream.source} not found when adding a filter"
```
fpacifici, line 232: You are not adding a filter.

sentry_streams/tests/adapters/arroyo/test_custom_strategies.py
```diff
@@ -0,0 +1,83 @@
...
+@pytest.fixture
+def forwarder() -> Forwarder:
+    return Forwarder(
+        route=Route(source="source", waypoints=["correct_branch"]),
+        produce_step=mock.Mock(spec=Produce),
+        next_step=mock.Mock(spec=ProcessingStrategy),
+    )
+
+
+def test_forwarder_submit(forwarder: Forwarder) -> None:
```
fpacifici, line 46: You are running one test only, you do not need a fixture.

sentry_streams/tests/adapters/arroyo/test_custom_strategies.py
```diff
@@ -0,0 +1,83 @@
...
+                offset=1,
+                timestamp=datetime(2025, 1, 1, 12, 0),
+            )
+        ),
+    }
+
+    forwarder.submit(messages["correct"])
+    forwarder._Forwarder__produce_step.submit.assert_called_once_with(expected_messages["correct"])  # type: ignore
+    forwarder.submit(messages["wrong"])
+    forwarder._Forwarder__next_step.submit.assert_called_once_with(expected_messages["wrong"]) == 1  # type: ignore
```
fpacifici, line 83: If you are testing a class you should not override the visibility. `__` methods are considered private. While python does not enforce this, it is not a good practice because the test becomes invalid if somebody refactors the class preserving the behavior. Testing only the public interface allows you to make changes to the internals while still using the existing test to validate your change. 

In this case run the assertions directly on the mock objects you passed to the strategy. You can access them as you created them.

Also please test the join, terminate, and close methods as well.

fpacifici (approved):
Good. 
Nothing critical. Please address the comments before merging, specifically the one on the topic config.

sentry_streams/sentry_streams/adapters/arroyo/adapter.py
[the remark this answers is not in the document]
bmckerry: > Do we need to introduce the override to make the router work ?

No, but all of our examples rely on the source topic being aliased to `logical-events` from `events`, so currently they don't work out of the box for the arroyo adapter.
I'll update our config to just have the actual topic be called `logical-events` as well and remove the aliasing here.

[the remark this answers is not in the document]
bmckerry: I've opened https://github.com/getsentry/streams/pull/84 to make it so our examples don't use topic aliasing

sentry_streams/sentry_streams/adapters/arroyo/custom_strategies.py
bmckerry: I was just referencing what `RunTask` does in `submit()`: https://github.com/getsentry/arroyo/blob/main/arroyo/processing/strategies/run_task.py#L51

I'll update to make a new message

sentry_streams/tests/adapters/arroyo/test_forwarder.py
```diff
@@ -0,0 +1,133 @@
...
+from arroyo.backends.kafka.consumer import KafkaPayload
+from arroyo.processing.strategies import Produce
+from arroyo.processing.strategies.abstract import ProcessingStrategy
+from arroyo.types import BrokerValue, FilteredPayload, Message, Partition, Topic
+
+from sentry_streams.adapters.arroyo.forwarder import Forwarder
+from sentry_streams.adapters.arroyo.routes import Route, RoutedValue
+
+
+# TODO: move make_msg into some kind of test utils folder as it's shared with other tests
```
ayirr7, line 14: would it be possbile to try out this TODO in this PR itself?

bmckerry: I'd like to tackle this in a separate PR as I'm not sure the best way to go about doing this atm

PR #84, opened 2025-03-26
ref: remove topics mapping from env config

bmckerry:
Our current way of configuring topics in `runnner.py` allows topics to be aliased to different  names, but the arroyo adapter has no knowledge of the topics config and doesn't work with aliased topics.

We'll reach a decision on this in the future, but for now keep our current structure but remove all the `logical-events` aliases for the `events` topic.

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,10 +29,10 @@ python sentry_streams/runner.py \
     -n test \
     -b localhost:9092 \
     -a arroyo \
-    sentry_streams/examples/transfomer.py
+    sentry_streams/examples/transformer.py
 ```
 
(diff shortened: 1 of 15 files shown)


ayirr7:
> but the arroyo adapter has no knowledge of the topics config and doesn't work with aliased topics.

Could you maybe link some code or show where this is happening? If the adapter doesn't have knowledge of the topics config, can we pass it through anyways? 

The main reason I'm asking is because we are going to add a deployment config soon anyways where all the adapters will haev to read from the topics config

ayirr7 (approved):
Just double checking that you were able to run all the examples on Flink and Arroyo adapter and they worked?

PR #87, opened 2025-03-31
Parse the pipeline deployment descriptor and generate k8s resources manifests

beezz:
https://github.com/getsentry/streaming-planning/issues/100

Using `deployment_config` from https://github.com/getsentry/streams/pull/77 this PR adds command that will generate k8s resources:

- deployment for each segment of pipeline configuration
- configmap for the configuration

To generate k8s resources using default templates:

```
python -m sentry_streams.k8s.generate  \
        --config sentry_streams/examples/sample_configs/deployment_config.yaml  \
        --image arroyo:latest  \
        --namespace pipeline-test
```

- [ ]  docker image ?
- [ ]  (default/required) resources requests/limits
- [ ] liveness and readiness probes
- [ ] update strategy
- [ ] deployment update on config change
- [ ] providing additional environment variables
- [ ] generic resources attributes overrides

cli support or custom templates:

- [ ] resources name templating
- [ ] resources labels templating
- [ ] annotations
- [ ] sidecars

### DIFF
--- a/sentry_streams/sentry_streams/k8s/generate.py
+++ b/sentry_streams/sentry_streams/k8s/generate.py
@@ -1,0 +1,180 @@
+import argparse
+import copy
+import importlib.resources
+import sys
+
+import yaml
+
+
+def _merge_labels(tpl_labels, seg_labels) -> dict:
+    """
+    Merge labels. Segment labels overrides the template labels.
+    """
+    return {**tpl_labels, **seg_labels}
+
+
+def _apply_namespace(k8s_resources, namespace: str | None = None):
+    for k8s_resource in k8s_resources:
+        k8s_resource["metadata"]["namespace"] = namespace
+
+
+def generate_configmap(*, config, configmap_template):
+    configmap = copy.deepcopy(configmap_template)
+    pipeline_name = config["pipeline"]["name"]
+    configmap["metadata"]["name"] = pipeline_name
+    common_labels = {"pipeline": pipeline_name}
+    configmap["metadata"]["labels"] = _merge_labels(
+        configmap["metadata"].get("labels", {}),
+        common_labels,
+    )
+    configmap["data"]["config"] = yaml.safe_dump(config)
+    return configmap
+
+
+def generate_deployments(*, config, deployment_template, container_name, image):
+    deployments = []
+    pipeline_name = config["pipeline"]["name"]
+    common_labels = {"pipeline": pipeline_name}
+
+    # configmap volume and mount
+    cm_volume = {
+        "name": f"{pipeline_name}-config",
+        "configMap": {"name": pipeline_name},
+    }
+    cm_volume_mount = {
+        "name": f"{pipeline_name}-config",
+        # TODO: Make mount/path configurable
+        "mountPath": f"/etc/{pipeline_name}-config",
+        "subPath": "config",
+        "readOnly": True,
+    }
+
+    for segid, segment in enumerate(config["pipeline"]["segments"]):
+        # TODO: sync with PR#77
+        # TODO: config as env variable as well?
+        labels = copy.deepcopy(common_labels)
+        labels["segment"] = segid
+
+        deployment = copy.deepcopy(deployment_template)
+
+        deployment["metadata"]["labels"] = _merge_labels(
+            deployment["metadata"].get("labels", {}),
+            labels,
+        )
+        deployment["spec"]["selector"]["matchLabels"] = _merge_labels(
+            deployment["spec"]["selector"].get("matchLabels", {}),
+            labels,
+        )
+        deployment["spec"]["template"]["metadata"]["labels"] = _merge_labels(
+            deployment["spec"]["template"]["metadata"]["labels"],
+            labels,
+        )
+
+        deployment["spec"]["replicas"] = segment["parallelism"]
+        deployment["metadata"]["name"] = f"{pipeline_name}-{segid}"
+
+        volumes = deployment["spec"].setdefault("volumes", [])
+        volumes.append(cm_volume)
+
+        for container in deployment["spec"]["template"]["spec"]["containers"]:
+            if container["name"] == container_name:
+                container["image"] = image
+
+                env = container.get("env", [])
+                env.append({"name": "SEGMENT_ID", "value": str(segid)})
+                container["env"] = env
+
+                volume_mounts = container.setdefault("volumeMounts", [])
+                volume_mounts.append(cm_volume_mount)
+
+                container["args"] = [
+                    "--config",
+                    cm_volume_mount["mountPath"],
+                ]
+
+        deployments.append(deployment)
+
+    return deployments
+
+
+def main() -> None:
+    parser = argparse.ArgumentParser(
+        description="Generate k8s resources from a sentry_streams deployment config."
+    )
+    parser.add_argument(
+        "--config",
+        type=argparse.FileType("r"),
+        help="Path to a deployment config file.",
+        required=True,
+    )
+    parser.add_argument(
+        "--deployment-template",
+        type=argparse.FileType("r"),
+        help="Path to a deployment template file.",
+        default=open(
+            importlib.resources.files("sentry_streams") / "k8s/templates/deployment.yaml",
+            "r",
+        ),
+    )
+    parser.add_argument(
+        "--configmap-template",
+        type=argparse.FileType("r"),
+        help="Path to a configmap template file.",
+        default=open(
+            importlib.resources.files("sentry_streams") / "k8s/templates/configmap.yaml",
+            "r",
+        ),
+    )
+    parser.add_argument(
+        "--output",
+        type=argparse.FileType("w"),
+        help="Output target file. Defaults to stdout.",
+        default=sys.stdout,
+    )
+    parser.add_argument(
+        "--container-name",
+        type=str,
+        help="Streams segment application container name.",
+        default="segment",
+    )
+    parser.add_argument(
+        "--image",
+        type=str,
+        help="Segment container image.",
+        required=True,
+    )
+    parser.add_argument(
+        "--namespace",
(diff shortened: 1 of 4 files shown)


beezz:
```
python -m sentry_streams.k8s.generate --config sentry_streams/examples/sample_configs/deployment_config.yaml --image arroyo:latest --namespace default

apiVersion: apps/v1
kind: Deployment
metadata:
  labels:
    pipeline: example-pipeline
    segment: 0
  name: example-pipeline-0
  namespace: default
spec:
  replicas: 2
  selector:
    matchLabels:
      pipeline: example-pipeline
      segment: 0
  template:
    metadata:
      labels:
        pipeline: example-pipeline
        segment: 0
    spec:
      containers:
      - args:
        - --config
        - /etc/example-pipeline-config
        command:
        - python
        - -m
        - sentry_streams.runner
        env:
        - name: SEGMENT_ID
          value: '0'
        image: arroyo:latest
        name: segment
        volumeMounts:
        - mountPath: /etc/example-pipeline-config
          name: example-pipeline-config
          readOnly: true
          subPath: config
  volumes:
  - configMap:
      name: example-pipeline
    name: example-pipeline-config
---
apiVersion: apps/v1
kind: Deployment
metadata:
  labels:
    pipeline: example-pipeline
    segment: 1
  name: example-pipeline-1
  namespace: default
spec:
  replicas: 3
  selector:
    matchLabels:
      pipeline: example-pipeline
      segment: 1
  template:
    metadata:
      labels:
        pipeline: example-pipeline
        segment: 1
    spec:
      containers:
      - args:
        - --config
        - /etc/example-pipeline-config
        command:
        - python
        - -m
        - sentry_streams.runner
        env:
        - name: SEGMENT_ID
          value: '1'
        image: arroyo:latest
        name: segment
        volumeMounts:
        - mountPath: /etc/example-pipeline-config
          name: example-pipeline-config
          readOnly: true
          subPath: config
  volumes:
  - configMap:
      name: example-pipeline
    name: example-pipeline-config
---
apiVersion: v1
data:
  config: "env:\n  topics:\n    events: events\n    transformed-events: transformed-events\n\
    \    transformed-events-2: transformed-events-2\npipeline:\n  name: example-pipeline\n\
    \  segments:\n  - parallelism: 2\n    steps_config:\n      myinput:\n        bootstrap_servers:\
    \ kafka:9093\n        starts_segment: true\n  - parallelism: 3\n    steps_config:\n\
    \      kafkasink:\n        bootstrap_servers: kafka:9093\n"
kind: ConfigMap
metadata:
  labels:
    pipeline: example-pipeline
  name: example-pipeline
  namespace: default
```

beezz:
```
python -m sentry_streams.k8s.generate --help
usage: generate.py [-h] --config CONFIG [--deployment-template DEPLOYMENT_TEMPLATE] [--configmap-template CONFIGMAP_TEMPLATE] [--output OUTPUT] [--container-name CONTAINER_NAME] --image IMAGE
                   [--namespace NAMESPACE]

Generate k8s resources from a sentry_streams deployment config.

options:
  -h, --help            show this help message and exit
  --config CONFIG       Path to a deployment config file.
  --deployment-template DEPLOYMENT_TEMPLATE
                        Path to a deployment template file.
  --configmap-template CONFIGMAP_TEMPLATE
                        Path to a configmap template file.
  --output OUTPUT       Output target file. Defaults to stdout.
  --container-name CONTAINER_NAME
                        Streams segment application container name.
  --image IMAGE         Segment container image.
  --namespace NAMESPACE
                        Namespace for deployment and configmap.

```

sentry_streams/sentry_streams/k8s/generate.py
```diff
@@ -0,0 +1,180 @@
...
+    )
+    parser.add_argument(
+        "--deployment-template",
+        type=argparse.FileType("r"),
+        help="Path to a deployment template file.",
+        default=open(
+            importlib.resources.files("sentry_streams") / "k8s/templates/deployment.yaml",
+            "r",
+        ),
+    )
```
fpacifici, line 118: like

fpacifici:
One high level question we will have to have at least a high level idea on. 
How does the generation of these resource fits in the SRE process to deploy Kuberentes resource ?

I see two options considering many pipelines will be running getsentry and, as of now, it seems SRE is pushing to keep the monolith intact in terms of k8s resources deployment.:
- sentry-kube becomes capable to deploy streaming pipeline. Imagine a sentry-infra-tool importing a package that contains the logic you provide here (I would like this code did not move into sentry-kube. Though we can publish a package sentry-kube can import). The pipelines are value files. We write a sentry-kube macro that triggers this code.
- Streaming pipeline are deployed with a different tool that directly generate manifests with what you are building and deploy via GoCD. Basically streaming pipeline behave as if they were their own services during deployment.

Doing it with an operator changes a bit the problem but not entirely. A resource still has to be deployed that references the getsentry pod template.

Did you already put any thought on this ?

sentry_streams/sentry_streams/k8s/generate.py
```diff
@@ -0,0 +1,197 @@
...
+        type=str,
+        help="Streams segment application container name.",
+        default="segment",
+    )
+    parser.add_argument(
+        "--image",
+        type=str,
+        help="Segment container image.",
+        required=True,
+    )
```
markstory, line 162: Would this be a getsentry/snuba image?

sentry_streams/sentry_streams/k8s/generate.py
```diff
@@ -0,0 +1,197 @@
...
+        config=config,
+        deployment_template=deployment_template,
+        image=args.image,
+        container_name=args.container_name,
+    ) + [
+        generate_configmap(
+            config=config,
+            configmap_template=configmap_template,
+        )
+    ]
```
fpacifici, line 184: I think you may have to generate and apply the configmaps before the deployments.

sentry_streams/sentry_streams/k8s/generate.py
```diff
@@ -0,0 +1,197 @@
...
+    for k8s_resource in k8s_resources:
+        k8s_resource["metadata"]["namespace"] = namespace
+
+
+def generate_configmap(
+    *, config: PipelineConfig, configmap_template: K8sConfigMapManifest
+) -> K8sConfigMapManifest:
+    configmap = copy.deepcopy(configmap_template)
+    pipeline_name = config["pipeline"]["name"]
+    configmap["metadata"]["name"] = pipeline_name
```
fpacifici, line 37: I think we may want to call the configmap `pipeline_<PIPELINE_NAME>` rather than `PIPELINE_NAME` as it has to be a unique name in the namespace.

sentry_streams/sentry_streams/k8s/operator/handlers.py
```diff
@@ -0,0 +1,140 @@
...
+        kopf.on.create("configmap", field="metadata.name", value=self.PIPELINES_CONFIGMAP)(
+            self.configure
+        )
+        kopf.on.update("configmap", field="metadata.name", value=self.PIPELINES_CONFIGMAP)(
+            self.configure
+        )
+
+
+streams_operator = StreamsOperator()
+streams_operator.register_handlers()
```
fpacifici, line 136: Does kopf require the operator to be defined at module level ?
If not can we have it instantiated in the main function or in a function called by main passing the config parameters into the constructor rather than transparently taking them from envvars in the constructor itself? It makes the modularization a bit better and unit tests easier.

sentry_streams/sentry_streams/k8s/operator/handlers.py
```diff
@@ -0,0 +1,140 @@
...
+            api_res = self.core_v1_client.read_namespaced_config_map(name=name, namespace=namespace)
+        except K8sApiException as exc:
+            if exc.status == 404:
+                # Create a new configmap.
+                self.logger.info(f"Create config map {name=}")
+                self.core_v1_client.create_namespaced_config_map(namespace=namespace, body=body)
+            else:
+                raise
+        # config-map already exists patch it
+        self.core_v1_client.patch_namespaced_config_map(name=name, namespace=namespace, body=body)
```
fpacifici, line 107: I'd add logs before the operator does any write operation.

sentry_streams/sentry_streams/k8s/operator/handlers.py
```diff
@@ -0,0 +1,140 @@
...
+        name, namespace = body["metadata"]["name"], body["metadata"]["namespace"]
+        try:
+            api_res = self.apps_v1_client.read_namespaced_deployment(name=name, namespace=namespace)
+        except K8sApiException as exc:
+            if exc.status == 404:
+                self.logger.info(f"Create deployment {name=}")
+                self.apps_v1_client.create_namespaced_deployment(namespace=namespace, body=body)
+            else:
+                raise
+        # do not patch the image
```
fpacifici, line 119: Is this because of GoCD ?

fpacifici (commented):
I think this goes in the right direction. 
Please see some comments inline. I think the key aspects to work on at this stage are:
- The CRD rather than using the configmap as we discussed off line
- move the code of the operator in its own package so `sentry_streams` does not have to depend on kops and kubernetes.

sentry_streams/sentry_streams/k8s/generate.py
markstory: How would we include sidecars, environment vars and additional configmaps that sentry will need?

sentry_streams/sentry_streams/k8s/operator/handlers.py
beezz: Yes, correct, so that the current way we deploy new images keeps working even for deployments managed by the operator.

sentry_streams/sentry_streams/k8s/generate.py
beezz: By using specific deployment template.

PR #88, opened 2025-03-31
feat: add explicit Broadcast step to pipeline graph

bmckerry:
Broadcast in flink doesn't require any processing as it works implicitly by just plugging a step into the `inputs` of mutliple downstream steps.
Broadcast in arroyo will need to do actual work on the adapter side, so we'll need an explicit `Broadcast()` step for that to work properly. This PR adds that step to the pipeline graph.
The `Broadcast()` step also has a routes list in the form of a list of `Branch()` much like the router step does.

### DIFF
--- a/sentry_streams/sentry_streams/examples/blq.py
+++ b/sentry_streams/sentry_streams/examples/blq.py
@@ -13,7 +13,7 @@ storage_branch = (
         "Send message to DBs",
         routes=[
             segment("sbc").sink("kafkasink", stream_name="transformed-events"),
-            segment("clickhouse").sink("kafkasink2", stream_name="transformed-events2"),
+            segment("clickhouse").sink("kafkasink2", stream_name="transformed-events-2"),
         ],
     )
 )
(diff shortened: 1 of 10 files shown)


sentry_streams/sentry_streams/examples/blq.py
```diff
@@ -13,7 +13,7 @@
         "Send message to DBs",
         routes=[
             segment("sbc").sink("kafkasink", stream_name="transformed-events"),
-            segment("clickhouse").sink("kafkasink2", stream_name="transformed-events2"),
+            segment("clickhouse").sink("kafkasink2", stream_name="transformed-events-2"),
```
bmckerry, line 16: I updated `transformed-events2` to `transformed-events-2` to keep the topic naming consistent with other examples/tests.

sentry_streams/sentry_streams/runner.py
```diff
@@ -34,6 +37,7 @@ def iterate_edges(p_graph: Pipeline, translator: RuntimeTranslator[StreamT, Stre
             step_streams[source_name] = source_streams[source_name]
 
         while step_streams:
+            logger.info(f"{step_streams=}")
```
fpacifici, line 40: remove ?

fpacifici (approved):
Please address the comment

PR #89, opened 2025-03-31
feat: add explicit broadcast step to flink

bmckerry:
Previously, we didn't have an explicit `Broadcast()` step as flink didn't require any processing in the adapter for broadcasting messages.
Arroyo won't work like this (we'll need to do processing in the adapter), so #88 added an explicit broadcast step to the graph. This adds that step to flink.


sentry_flink/tests/test_pipeline.py
```diff
@@ -224,6 +225,29 @@ def basic_filter() -> tuple[Pipeline, MutableMapping[str, list[dict[str, Any]]]]
     return (pipeline, expected)
 
 
+def basic_broadcast() -> tuple[Pipeline, MutableMapping[str, list[dict[str, Any]]]]:
+    pipeline = streaming_source(name="mysource", stream_name="events").broadcast(
```
bmckerry, line 229: in a future PR I'll update all the test pipelines here to use the chaining syntax

PR #91, opened 2025-04-01
feat: add arroyo broadcast step

bmckerry:
This PR implements Broadcast in the arroyo adapter.

Broadcast works by copying a given message once for each downstream branch, updating the Route of each copy to correspond to one of the branches. This is done using the arroyo [unfold](https://getsentry.github.io/arroyo/strategies/unfold.html) strategy.


sentry_streams/sentry_streams/adapters/arroyo/steps.py
```diff
@@ -123,6 +130,53 @@ def transformer(
...
+            else:
+                return [
+                    Value(
+                        payload=RoutedValue(
+                            route=Route(
+                                source=payload.route.source,
+                                waypoints=cast(
+                                    MutableSequence[str],
+                                    cast(List[str], payload.route.waypoints)
+                                    + [downstream_branch.name],
```
bmckerry, line 164: In order to keep `waypoints` as a `MutableSequence` and not a `List`, I need to do this `cast` nonsense in order to add the downstream branch step name to the message copy's waypoints. Very open to better ways of doing this (alternatively we can make waypoints just be a `List` which means I can use the `+` operator natively).

sentry_streams/sentry_streams/examples/blq.py
```diff
@@ -23,7 +23,7 @@
     .apply("dump_msg_delayed", Map(json_dump_message))
     .sink(
         "kafkasink3",
-        stream_name="transformed-events3",
+        stream_name="transformed-events-3",
```
bmckerry, line 26: updated so all of our examples/tests match re: topic name

sentry_streams/tests/adapters/arroyo/conftest.py
```diff
@@ -28,84 +31,75 @@ def broker() -> LocalBroker[KafkaPayload]:
 
 @pytest.fixture
 def pipeline() -> Pipeline:
```
bmckerry, line 33: updated all pipelines in this file to use the chaining syntax

sentry_streams/sentry_streams/adapters/arroyo/steps.py
```diff
@@ -123,6 +130,53 @@ def transformer(
...
+    each copy having a Route matching one of the downstream branches.
+    """
+
+    pipeline_step: Broadcast
+
+    def build(
+        self, next: ProcessingStrategy[Union[FilteredPayload, RoutedValue]], commit: Commit
+    ) -> ProcessingStrategy[Union[FilteredPayload, RoutedValue]]:
+        def generate_message_copies(
+            payload: RoutedValue,
```
bmckerry, line 146: Unfortunately, the generator function used by `unfold` just gets passed the message Value's payload (in our case the `RoutedValue`) and not the Value itself. This means we lose access to the `timestamp` field of the original message, and any messages which pass thru a Broadcast step have a timestamp of `None`.

I can update the function here to use the current time as the timestamp on the message copies, not sure if that's desirable at all.

bmckerry: from @lynnagara: we could also change the unfold function to take the whole message, but would need to audit every use of `unfold` across sentry atm and update as this would be a breaking change

fpacifici (changes requested):
Please see the comments inline

sentry_streams/tests/adapters/arroyo/test_broadcaster.py
```diff
@@ -0,0 +1,114 @@
...
+
+from arroyo.processing.strategies.abstract import ProcessingStrategy
+from arroyo.types import FilteredPayload, Message, Partition, Topic, Value
+
+from sentry_streams.adapters.arroyo.broadcaster import Broadcaster
+from sentry_streams.adapters.arroyo.routes import Route, RoutedValue
+
+
+# TODO: move make_value_msg into some kind of test utils folder as it's shared with other tests
+def make_value_msg(payload: Any, route: Route, offset: int) -> Message[Any]:
```
bmckerry, line 14: we have these `make_msg` functions repeated across all of our arroyo test folders - I'd like to tackle moving them to pytest fixtures tomorrow as I'm not exactly sure how I'd go about doing this and will probably need to restructure a bunch of our arroyo tests

sentry_streams/tests/adapters/arroyo/test_steps.py
```diff
@@ -32,13 +53,37 @@ def make_msg(payload: Any, route: Route, offset: int) -> Message[Any]:
...
                 payload=RoutedValue(route=route, payload=payload),
-                partition=Partition(Topic("test_topic"), 0),
+                partition=TEST_PARTITION,
                 offset=offset,
                 timestamp=datetime(2025, 1, 1, 12, 0),
             )
         )
 
 
+def make_value_msg(payload: Any, route: Route, offset: int) -> Message[Any]:
```
ayirr7, line 63: crazy enough, I added this to main right before you yesterday 

seems identical to the current version in main, but just wanted to point that out in case theres a conflict

sentry_streams/sentry_streams/adapters/arroyo/adapter.py
```diff
@@ -232,7 +233,20 @@ def broadcast(
...
+        assert (
+            stream.source in self.__consumers
+        ), f"Stream starting at source {stream.source} not found when adding a broadcast step"
+        self.__consumers[stream.source].add_step(BroadcastStep(route=stream, pipeline_step=step))
+
+        return {
+            route.name: Route(
+                source=stream.source,
+                waypoints=cast(
+                    MutableSequence[str], cast(List[str], stream.waypoints) + [route.name]
```
ayirr7, line 245: Just for my understanding -- what's the meaning behind adding both the route and the route's waypoints?

sentry_streams/tests/adapters/arroyo/test_broadcaster.py
```diff
@@ -0,0 +1,114 @@
...
+        ),
+        call.submit(
+            make_value_msg(
+                payload="test-payload",
+                route=Route(source="source", waypoints=["branch_2"]),
+                offset=0,
+            )
+        ),
+    ]
+    print(expected_calls)
```
ayirr7, line 67: nit; let's remove any of these

sentry_streams/sentry_streams/adapters/arroyo/steps.py
```diff
@@ -123,6 +130,25 @@ def transformer(
...
+class BroadcastStep(ArroyoStep):
+    """
+    BroadcastStep forwards a copy of an incoming  message downstream for each downstream branch,
+    each copy having a Route matching one of the downstream branches.
+    """
+
+    pipeline_step: Broadcast
+
+    def build(
+        self, next: ProcessingStrategy[Union[FilteredPayload, RoutedValue]], commit: Commit
```
ayirr7, line 143: I noticed that this is main already but, re: commit

What's the meaning behind having each Arroyo step take in a Commit? Is the idea that the commit policy can be different per step?

ayirr7 (approved):
Overall looks good to me, maybe wait for @fpacifici's approval if there are specific comments

sentry_streams/sentry_streams/adapters/arroyo/adapter.py
```diff
@@ -232,7 +233,20 @@ def broadcast(
...
         """
-        raise NotImplementedError
+        assert (
+            stream.source in self.__consumers
+        ), f"Stream starting at source {stream.source} not found when adding a broadcast step"
+        self.__consumers[stream.source].add_step(BroadcastStep(route=stream, pipeline_step=step))
+
+        return {
+            route.name: Route(
+                source=stream.source,
```
ayirr7, line 243: Just a general question but is my understanding correct that we track the Source as part of the Route so that we can commit offsets properly?

fpacifici: The source in the route was not really for the consumer at runtime but for the adapter to be able to plug steps to the right stream.

sentry_streams/sentry_streams/adapters/arroyo/adapter.py
fpacifici, line 245: No need to cast
```
waypoints = [*stream.waypoints, route.name]
```

sentry_streams/sentry_streams/adapters/arroyo/broadcaster.py
```diff
@@ -0,0 +1,57 @@
...
+                msg_copy = Message(
+                    Value(
+                        committable=message.value.committable,
+                        timestamp=message.value.timestamp,
+                        payload=RoutedValue(
+                            route=Route(
+                                source=message_payload.route.source,
+                                waypoints=[*message_payload.route.waypoints, *[branch]],
+                            ),
+                            payload=message_payload.payload,
```
fpacifici, line 39: You'll have to make a copy of the message. The downstream branches can mutate it.
https://docs.python.org/3/library/copy.html#copy.deepcopy

sentry_streams/sentry_streams/adapters/arroyo/broadcaster.py
```diff
@@ -0,0 +1,57 @@
...
+                        ),
+                    )
+                )
+                self.__next_step.submit(msg_copy)
```
fpacifici, line 43: This can raise a MessageRejected exception. 
As you are multiplying the messages downstream you do not want to send the message twice if it succeeds on one route but fails on another. 

Arroyo works this way in this scenario:
- arroyo calls submit on your strategy
- you do stuff and call submit on downstream steps.
- any of those call can result in a MessageRejected exception, which means backpressure. 
- If you re-raise the exception arroyo will retry to send you the same message till it succeeds. 

In most cases where each step has a 0..1 output cardinality simply re-raising the esception is enough.
In this case you are calling multiple submit to downstream steps. Any of them can fail while another can succeed. 


What you can do is to keep track in this class of the last message sent to each of the downstream route successfully. If you raise a MessageRejected exception, when arroyo calls submit again you do not send downstream the message if it was already successful.

sentry_streams/tests/adapters/arroyo/test_broadcaster.py
fpacifici: Can't you just create a module in the tests directory and import it via 
```
from tests.your_module..... import ...
```

conftest is for fixture, this is not a fixture and other customizations of pytest.

sentry_streams/tests/adapters/arroyo/test_broadcaster.py
```diff
@@ -0,0 +1,114 @@
...
+        return Message(
+            Value(
+                payload=RoutedValue(route=route, payload=payload),
+                committable={Partition(Topic("test_topic"), 0): offset},
+                timestamp=timestamp,
+            )
+        )
+
+
+def test_submit() -> None:
```
fpacifici, line 39: Please test the negative cases where you send: FilteredMessage and messages for another route.
Also please test the case where the downstream step raises MessageRejected on the second route after succeeding the first time. Ensure the message is forwarded only once per route.

When writing unit tests it is all the more important to test teh corner cases as they are the easier ones to break without noticing.

fpacifici (commented):
Please see the comment on MessageRejected. This implementaiton risks to duplicate messages.

sentry_streams/tests/adapters/arroyo/test_broadcaster.py
bmckerry: whoops, I was getting import errors because I was trying to import `from sentry_streams.tests.adapters...` instead of `from tests.adapters....`

Will move them to a helper file now, thank you

sentry_streams/sentry_streams/adapters/arroyo/adapter.py
bmckerry: In this loop `route` corresponds to a `Branch` downstream of the broadcast step - for each downstream branch, we need to update the `waypoints` of that branch's `Route` since we use the `Route` to determine if messages should be processed by that step or not. This updated `Route` will get propagated to all steps downstream from each branch.

Let me know if that doesn't answer your question, happy to explain more

sentry_streams/sentry_streams/adapters/arroyo/steps.py
bmckerry: The commit policy is required by the `SinkStep` for the producer that step creates: https://github.com/getsentry/streams/blob/main/sentry_streams/sentry_streams/adapters/arroyo/steps.py#L185

Since all steps inherit from `ArroyoStep`, to add `commit` to `SinkStep` I also had to update `ArroyoStep` to pass commit, which means I also had to update everything that inherits from `ArroyoStep` (even though they don't use this param). Very open to better ways of doing this, I can update in a future PR.

sentry_streams/sentry_streams/adapters/arroyo/broadcaster.py
```diff
@@ -0,0 +1,89 @@
...
+                                waypoints=[*copy_payload.route.waypoints, branch],
+                            ),
+                            payload=copy_payload.payload,
+                        ),
+                    )
+                )
+                try:
+                    self.__next_step.submit(routed_copy)
+                except MessageRejected:
+                    self.__pending.append(routed_copy)
```
fpacifici, line 66: You have to raise the MessageRejected exception otherwise Arroyo would interpret the submit as a success and send you the next message.

sentry_streams/sentry_streams/adapters/arroyo/broadcaster.py
```diff
@@ -0,0 +1,89 @@
...
+        self.__pending: Deque[Message[RoutedValue]] = deque()
+
+    def __flush_pending(self) -> None:
+        while self.__pending:
+            try:
+                message = self.__pending[0]
+                self.__next_step.submit(message)
+                self.__pending.popleft()
+            except MessageRejected:
+                pass
```
fpacifici, line 38: This code will end in an infinite loop if the downstream step is saturated without a way to make progress.
Imagine the downstream step is at capacity.
You will start trying to submit the message till the downstream step keep raising MessageRejected.
Though, as you would never call next.poll in this loop, the downstream step will never make progress and keep raising the exception.

sentry_streams/tests/adapters/arroyo/test_broadcaster.py
```diff
@@ -0,0 +1,201 @@
...
+        call.submit(
+            make_value_msg(
+                payload="test-payload",
+                route=Route(source="source", waypoints=["branch_2"]),
+                offset=0,
+            )
+        ),
+    ]
+
+    broadcaster.submit(message)
```
fpacifici, line 152: The contract is violated here. Your step should not swallow the MessageRejected exception otherwise Arroyo will not know it is in a backpressure state. It will only know that at the following message.

sentry_streams/tests/adapters/arroyo/test_broadcaster.py
```diff
@@ -0,0 +1,201 @@
...
+
+    # stop raising MessageRejected
+    next_step.submit.side_effect = None
```
fpacifici, line 155: Please assert the calls are the right one before changing the side effect or you are missing test cases.

sentry_streams/sentry_streams/adapters/arroyo/broadcaster.py
```diff
@@ -0,0 +1,130 @@
...
+                            ),
+                            payload=copy_payload.payload,
+                        ),
+                    )
+                )
+                try:
+                    self.__handle_submit(routed_copy)
+                except MessageRejected:
+                    # Raise a single MessageRejected at the end of the loop
+                    raise_message_rejected = True
```
fpacifici, line 101: You do not need to do this.
Just raise the exception at the first failure.
So you can only have one failed message

fpacifici (approved):

PR #99, opened 2025-04-14
ref: Type out chains and steps

ayirr7:
Does the following:

- Links types of different steps of an ExtensibleChain
- Provides a simple Message class to carry message metadata through the pipeline. This allows for decoupling certain steps, like parsing/deserialization from the stream source step. 
- Basic message decoders/encoders provided out of the box in the API (JSON, protobuf)
- Basic header filtering in the source
- Schema validation and transforming raw bytes into a schema-enforced message type
- General type checks
- Unit tests cleanup

Still to do:

- Improve a bunch of naming (of interfaces, of files, etc.)
- Provide full support for Protobuf

### DIFF
--- a/sentry_streams/sentry_streams/adapters/arroyo/consumer.py
+++ b/sentry_streams/sentry_streams/adapters/arroyo/consumer.py
@@ -1,19 +1,15 @@
 import logging
 from dataclasses import dataclass, field
-from typing import Any, Mapping, MutableSequence
+from typing import Any, Mapping, MutableSequence, Optional, Tuple, Union
 
-from arroyo.backends.kafka.consumer import KafkaPayload
+from arroyo.backends.kafka.consumer import Headers, KafkaPayload
 from arroyo.processing.strategies import CommitOffsets
 from arroyo.processing.strategies.abstract import (
     ProcessingStrategy,
     ProcessingStrategyFactory,
 )
 from arroyo.processing.strategies.run_task import RunTask
-from arroyo.types import (
-    Commit,
-    Message,
-    Partition,
-)
+from arroyo.types import Commit, FilteredPayload, Message, Partition
 
 from sentry_streams.adapters.arroyo.routes import Route, RoutedValue
 from sentry_streams.adapters.arroyo.steps import ArroyoStep
@@ -41,6 +37,7 @@ class ArroyoConsumer:
     """
 
     source: str
+    header_filter: Optional[Tuple[str, bytes]] = None
     steps: MutableSequence[ArroyoStep] = field(default_factory=list)
 
     def add_step(self, step: ArroyoStep) -> None:
@@ -59,9 +56,20 @@ class ArroyoConsumer:
         follow.
         """
 
-        def add_route(message: Message[KafkaPayload]) -> RoutedValue:
-            value = message.payload.value
-            return RoutedValue(route=Route(source=self.source, waypoints=[]), payload=value)
+        def add_route(message: Message[KafkaPayload]) -> Union[FilteredPayload, RoutedValue]:
+            filtered = True
+            if self.header_filter:
+                exp_k, exp_v = self.header_filter
+                headers: Headers = message.payload.headers
+                for k, v in headers:
+                    if k == exp_k and v == exp_v:
+                        filtered = False
+
+            if not filtered:
+                value = message.payload.value
+                return RoutedValue(route=Route(source=self.source, waypoints=[]), payload=value)
+            else:
+                return FilteredPayload()
 
         strategy: ProcessingStrategy[Any] = CommitOffsets(commit)
         for step in reversed(self.steps):
--- a/sentry_streams/sentry_streams/adapters/arroyo/consumer.py
+++ b/sentry_streams/sentry_streams/adapters/arroyo/consumer.py
@@ -1,4 +1,5 @@
 import logging
+import time
 from dataclasses import dataclass, field
 from typing import Any, Mapping, MutableSequence, Optional, Tuple, Union
 
@@ -62,8 +63,8 @@ class ArroyoConsumer:
 
         def add_route(message: Message[KafkaPayload]) -> Union[FilteredPayload, RoutedValue]:
             filtered = False
+            headers: Headers = message.payload.headers
             if self.header_filter:
-                headers: Headers = message.payload.headers
                 if self.header_filter not in headers:
                     filtered = True
 
@@ -76,9 +77,12 @@ class ArroyoConsumer:
                 except Exception:
                     raise ValueError(f"Kafka topic {self.stream_name} has no associated schema")
 
+                now = time.time()
                 return RoutedValue(
                     route=Route(source=self.source, waypoints=[]),
-                    payload=StreamsMessage(schema=schema, payload=value),
+                    payload=StreamsMessage(
+                        payload=value, headers=headers, timestamp=now, schema=schema
+                    ),
                 )
 
         strategy: ProcessingStrategy[Any] = CommitOffsets(commit)
--- a/sentry_streams/sentry_streams/adapters/arroyo/consumer.py
+++ b/sentry_streams/sentry_streams/adapters/arroyo/consumer.py
@@ -11,7 +11,6 @@ from arroyo.processing.strategies.abstract import (
 )
 from arroyo.processing.strategies.run_task import RunTask
 from arroyo.types import Commit, FilteredPayload, Message, Partition
-from sentry_kafka_schemas import get_codec
 from sentry_kafka_schemas.codecs import Codec
 
 from sentry_streams.adapters.arroyo.routes import Route, RoutedValue
@@ -42,6 +41,7 @@ class ArroyoConsumer:
 
     source: str
     stream_name: str
+    schema: Codec[Any]
     header_filter: Optional[Tuple[str, bytes]] = None
     steps: MutableSequence[ArroyoStep] = field(default_factory=list)
 
@@ -68,15 +68,11 @@ class ArroyoConsumer:
 
             broker_timestamp = message.timestamp.timestamp() if message.timestamp else time.time()
             value = message.payload.value
-            try:
-                schema: Codec[Any] = get_codec(self.stream_name)
-            except Exception:
-                raise ValueError(f"Kafka topic {self.stream_name} has no associated schema")
 
             return RoutedValue(
                 route=Route(source=self.source, waypoints=[]),
                 payload=StreamsMessage(
-                    payload=value, headers=headers, timestamp=broker_timestamp, schema=schema
+                    payload=value, headers=headers, timestamp=broker_timestamp, schema=self.schema
                 ),
             )
 
--- a/sentry_streams/sentry_streams/adapters/arroyo/consumer.py
+++ b/sentry_streams/sentry_streams/adapters/arroyo/consumer.py
@@ -10,9 +10,12 @@ from arroyo.processing.strategies.abstract import (
 )
 from arroyo.processing.strategies.run_task import RunTask
 from arroyo.types import Commit, FilteredPayload, Message, Partition
+from sentry_kafka_schemas import get_codec
+from sentry_kafka_schemas.codecs import Codec
 
 from sentry_streams.adapters.arroyo.routes import Route, RoutedValue
 from sentry_streams.adapters.arroyo.steps import ArroyoStep
+from sentry_streams.pipeline.chain import Message as StreamsMessage
 
 logger = logging.getLogger(__name__)
 
@@ -67,7 +70,11 @@ class ArroyoConsumer:
 
             if not filtered:
                 value = message.payload.value
-                return RoutedValue(route=Route(source=self.source, waypoints=[]), payload=value)
+                schema: Codec[Any] = get_codec(self.source)
+                return RoutedValue(
+                    route=Route(source=self.source, waypoints=[]),
+                    payload=StreamsMessage(schema=schema, payload=value),
+                )
             else:
                 return FilteredPayload()
 
--- a/sentry_streams/sentry_streams/adapters/arroyo/consumer.py
+++ b/sentry_streams/sentry_streams/adapters/arroyo/consumer.py
@@ -62,28 +62,23 @@ class ArroyoConsumer:
         """
 
         def add_route(message: Message[KafkaPayload]) -> Union[FilteredPayload, RoutedValue]:
-            filtered = False
             headers: Headers = message.payload.headers
-            if self.header_filter:
-                if self.header_filter not in headers:
-                    filtered = True
-
-            if filtered:
+            if self.header_filter and self.header_filter not in headers:
                 return FilteredPayload()
-            else:
-                value = message.payload.value
-                try:
-                    schema: Codec[Any] = get_codec(self.stream_name)
-                except Exception:
-                    raise ValueError(f"Kafka topic {self.stream_name} has no associated schema")
 
-                now = time.time()
-                return RoutedValue(
-                    route=Route(source=self.source, waypoints=[]),
-                    payload=StreamsMessage(
-                        payload=value, headers=headers, timestamp=now, schema=schema
-                    ),
-                )
+            broker_timestamp = message.timestamp.timestamp() if message.timestamp else time.time()
+            value = message.payload.value
+            try:
+                schema: Codec[Any] = get_codec(self.stream_name)
+            except Exception:
+                raise ValueError(f"Kafka topic {self.stream_name} has no associated schema")
+
+            return RoutedValue(
+                route=Route(source=self.source, waypoints=[]),
+                payload=StreamsMessage(
+                    payload=value, headers=headers, timestamp=broker_timestamp, schema=schema
+                ),
+            )
 
         strategy: ProcessingStrategy[Any] = CommitOffsets(commit)
         for step in reversed(self.steps):
--- a/sentry_streams/sentry_streams/adapters/arroyo/consumer.py
+++ b/sentry_streams/sentry_streams/adapters/arroyo/consumer.py
@@ -69,7 +69,11 @@ class ArroyoConsumer:
 
             if not filtered:
                 value = message.payload.value
-                schema: Codec[Any] = get_codec(self.stream_name)
+                try:
+                    schema: Codec[Any] = get_codec(self.stream_name)
+                except Exception:
+                    raise ValueError(f"Kafka topic {self.stream_name} has no associated schema")
+
                 return RoutedValue(
                     route=Route(source=self.source, waypoints=[]),
                     payload=StreamsMessage(schema=schema, payload=value),
--- a/sentry_streams/sentry_streams/adapters/arroyo/consumer.py
+++ b/sentry_streams/sentry_streams/adapters/arroyo/consumer.py
@@ -15,7 +15,7 @@ from sentry_kafka_schemas.codecs import Codec
 
 from sentry_streams.adapters.arroyo.routes import Route, RoutedValue
 from sentry_streams.adapters.arroyo.steps import ArroyoStep
-from sentry_streams.pipeline.chain import Message as StreamsMessage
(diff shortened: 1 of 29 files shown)


sentry_streams/sentry_streams/examples/transformer.py
```diff
@@ -58,12 +59,12 @@ def merge(self, other: Self) -> Self:
         name="myinput",
         stream_name="events",
     )
-    .apply("mymap", Map(function=parse))
+    .apply("parser", Parser(deserializer=json_parser))
```
ayirr7, line 62: This gives the user flexibility to defer deserializng the message, but I'd like to maybe find a way to do it where the platform figures out where to do the parsing based on some info the user provides.

sentry_streams/sentry_streams/pipeline/msg_parser.py
```diff
@@ -0,0 +1,15 @@
+from json import JSONDecodeError, dumps, loads
+from typing import Any, Mapping, cast
+
+
+def json_parser(msg: bytes) -> Mapping[str, Any]:
```
ayirr7, line 5: These are supposed to be the standard serializer/deserializer functions that the user will plug in. Chose them to be functions so that the typing inference works out more seamlessly between steps in a chain, but up to change

sentry_streams/sentry_streams/pipeline/chain.py
```diff
@@ -39,12 +45,22 @@
...
+
+
 TRoute = TypeVar("TRoute")
 
 TIn = TypeVar("TIn")
 TOut = TypeVar("TOut")
 
 
+# a message with a generic payload
+class Message(Generic[TIn]):
```
ayirr7, line 58: I am trying to intentionally keep `Message` as bare-bones as possible. For example, I don't think the user should have to think about:
- headers after the source step
- partition key
- offset management

However, they can provide a schema for messages and any additional info. e.g. for the generic metrics pipeline, maybe they annotate each message with the metric type. 

Do let me know if something important is missing though.

sentry_streams/sentry_streams/examples/transformer.py
```diff
@@ -38,12 +39,13 @@ def __init__(self) -> None:
         self.batch: MutableSequence[Any] = []
 
     def add(self, value: Any) -> Self:
```
ayirr7, line 41: Message[] will need to be become part of the type annotations within Accumulator class definition, etc. I'm working on it

ayirr7:
Got some offline reviews from @untitaker 

Currently working on:
- Adding schema validation to the `Parser` step and updating an example to use that
- Seeing if types for certain steps should be more narrow (i.e. what the Serializer outputs)

sentry_streams/sentry_streams/pipeline/message.py
```diff
@@ -0,0 +1,27 @@
...
+    TypeVar,
+)
+
+from sentry_kafka_schemas.codecs import Codec
+
+TIn = TypeVar("TIn")
+
+
+# a message with a generic payload
+class Message(Generic[TIn]):
```
ayirr7, line 17: I am trying to intentionally keep Message as bare-bones as possible. For example, I don't think the user should have to think about:

headers after the source step
partition key
offset management
However, they can provide a schema for messages and any additional info. e.g. for the generic metrics pipeline, maybe they annotate each message with the metric type.

Do let me know if something important is missing though.

sentry_streams/sentry_streams/adapters/arroyo/consumer.py
```diff
@@ -59,9 +60,26 @@ def build_strategy(self, commit: Commit) -> ProcessingStrategy[Any]:
...
+            if not filtered:
+                value = message.payload.value
+                try:
+                    schema: Codec[Any] = get_codec(self.stream_name)
+                except Exception:
+                    raise ValueError(f"Kafka topic {self.stream_name} has no associated schema")
+
+                return RoutedValue(
+                    route=Route(source=self.source, waypoints=[]),
+                    payload=StreamsMessage(schema=schema, payload=value),
```
ayirr7, line 79: This identifies the schema of the messages based on the topic it comes from. I chose to do it this way, and wrap it in this `Message` so that we can do flexible parsing/deserialization of messages. This way, instead of baking it into the Source, we can basically do parsing whenever in the pipeline using the `Parser()` step.

sentry_streams/sentry_streams/examples/batching.py
```diff
@@ -1,32 +1,33 @@
...
+# TODO: Figure out why the concrete type of InputType is not showing up in the type hint of chain1
+chain1 = pipeline.apply("parser", Parser(msg_type=IngestMetric, deserializer=json_parser)).apply(
+    "mybatch", Batch(batch_size=5)
+)  # User simply provides the batch size
+
+chain2 = chain1.apply(
+    "myunbatch",
+    FlatMap(
+        function=cast(
+            Union[Callable[[Message[MutableSequence[IngestMetric]]], Message[IngestMetric]], str],
```
ayirr7, line 25: Because `unbatch` is a function with generic arguments, the user unfortunately has to explicitly cast it to the concrete type in their pipeline. This shouldn't be difficult because the incoming `ExtensibleChain` already has concrete type hints. 

Obviously this specific `cast()` can also be avoided if users write and use a custom unbatcher.

sentry_streams/sentry_streams/examples/transformer.py
```diff
@@ -53,17 +47,16 @@ def merge(self, other: Self) -> Self:
...
-    .apply("myreduce", Reducer(reduce_window, TransformerBatch))
-    .apply("serializer", Map(function=serialize_msg))
-    .sink(
-        "kafkasink2",
-        stream_name="transformed-events",
-    )  # flush the batches to the Sink
-)
+pipeline = streaming_source(name="myinput", stream_name="ingest-metrics")
+
+chain2 = pipeline.apply("parser", Parser(msg_type=IngestMetric, deserializer=json_parser))
```
ayirr7, line 52: I split up the pipeline like this because you can see the inferred type hints in an editor

sentry_streams/sentry_streams/examples/transformer.py
```diff
@@ -53,17 +47,24 @@ def merge(self, other: Self) -> Self:
...
+chain2 = chain1.apply(
+    "custom_batcher", Reducer(reduce_window, TransformerBatch)
+)  # ExtensibleChain[Message[MutableSequence[IngestMetric]]]
+
+chain3 = chain2.apply(
+    "serializer",
+    Serializer(serializer=json_serializer),
+)  # ExtensibleChain[bytes]
+
+chain4 = chain3.sink(
```
ayirr7, line 67: I split up the pipeline like this because you can see the inferred type hints of each chain in a code editor

sentry_streams/sentry_streams/examples/transformer.py
```diff
@@ -53,17 +47,27 @@ def merge(self, other: Self) -> Self:
...
-    )  # flush the batches to the Sink
-)
+pipeline = streaming_source(
+    name="myinput", stream_name="ingest-metrics"
+)  # ExtensibleChain[Message[bytes]]
+
+chain1 = pipeline.apply(
+    "parser",
+    Parser(
+        msg_type=IngestMetric, deserializer=msg_parser
```
ayirr7, line 57: I don't think we need the `deserializer=msg_parser` specified like this because it's meant to be generic and that is the only deserializer

sentry_streams/sentry_streams/adapters/arroyo/consumer.py
```diff
@@ -59,9 +60,26 @@ def build_strategy(self, commit: Commit) -> ProcessingStrategy[Any]:
...
-            return RoutedValue(route=Route(source=self.source, waypoints=[]), payload=value)
+        def add_route(message: Message[KafkaPayload]) -> Union[FilteredPayload, RoutedValue]:
+            filtered = False
+            if self.header_filter:
+                headers: Headers = message.payload.headers
+                if self.header_filter not in headers:
+                    filtered = True
+
+            if filtered:
+                return FilteredPayload()
```
fpacifici, line 71:
```
if self.header_filter:
    headers: Headers = message.payload.headers
    if self.header_filter not in headers:
           return FilteredPayload()
```

sentry_streams/sentry_streams/adapters/arroyo/consumer.py
```diff
@@ -59,9 +60,26 @@ def build_strategy(self, commit: Commit) -> ProcessingStrategy[Any]:
...
+            else:
+                value = message.payload.value
+                try:
+                    schema: Codec[Any] = get_codec(self.stream_name)
```
fpacifici, line 75: I think you should not need to run the `get_codec` function for every message. I'd get the instance of the schema only once and append it as a flyweight object to all messages.

sentry_streams/sentry_streams/examples/batching.py
fpacifici: I wonder whether it would be easier if we had  `Batch` and `Unbatch` appliers without having to go through the FlatMap.

sentry_streams/sentry_streams/pipeline/msg_parser.py
```diff
@@ -0,0 +1,31 @@
+from typing import Any
+
+from sentry_streams.pipeline.message import Message
```
fpacifici, line 3: Unit tests please ensuring it works for protobuf as well

fpacifici (commented):
I think it goes in a good direction. 

See the comments in line for details but my main feedback is to be more strict on the message interface. Specifically:
- I would separate the concept of parsed message (only headers) and serialized (pre-parsing and post serialization) message where the payload is available already parsed. We will figure out later whether to relax this constraint. This would also make it clear that serialization has to happen before a sink.
- Maybe later, introduce a difference in terms of types between serializable messages (where we can break segments) and any other type where we cannot break segments. 

Also, what happens today for invalid messages? We will have to add support for the DLQ. It is ok to do it separately, but please call it out in the parser.

sentry_streams/sentry_streams/pipeline/message.py
markstory: > Expose headers and separate them between internal ones managed by platform, mutable by the platform and the application ones: readonly and have to be present at the source. We will figure out later how to provide a more flexible set of application headers

With tasks, we have headers as as simple mutable map. Separating the headers out seems like additional complexity we could avoid. Our SDKs use message headers to pass trace_id and trace baggage which I assume you'll want for streaming as well.

sentry_streams/sentry_streams/pipeline/msg_parser.py
markstory: You'll need a protobuf message type published to do that. If you need a hand with sentry-protos let me know.

sentry_streams/sentry_streams/pipeline/message.py
fpacifici: for the time being: add headers as a map, propagate in all steps, leave empty in reduce.
Add timestamp.

sentry_streams/sentry_streams/examples/batching.py
fpacifici: out of the scope for now.

sentry_streams/sentry_streams/pipeline/chain.py
```diff
@@ -246,25 +286,25 @@ def sink(self, name: str, stream_name: str) -> Chain:
...
     """
-    pipeline: ExtensibleChain = ExtensibleChain(name)
+    pipeline: ExtensibleChain[Message[TIn]] = ExtensibleChain(name)
     pipeline._add_start(Branch(name=name, ctx=pipeline))
     return pipeline
 
 
-def streaming_source(name: str, stream_name: str) -> ExtensibleChain:
+def streaming_source(
+    name: str, stream_name: str, header_filter: Optional[Tuple[str, bytes]] = None
```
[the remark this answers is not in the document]
fpacifici: solved.

sentry_streams/sentry_streams/pipeline/chain.py
```diff
@@ -69,34 +74,37 @@ def build_step(self, name: str, ctx: Pipeline, previous: Step) -> Step:
 
 
 @dataclass
-class Map(Applier[TIn, TOut], Generic[TIn, TOut]):
-    function: Union[Callable[[TIn], TOut], str]
+class Map(Applier[Message[TIn], Message[TOut]], Generic[TIn, TOut]):
+    function: Union[Callable[[Message[TIn]], Message[TOut]], str]
```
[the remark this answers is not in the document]
ayirr7: This is addressed now

sentry_streams/sentry_streams/adapters/arroyo/consumer.py
```diff
@@ -59,9 +61,29 @@ def build_strategy(self, commit: Commit) -> ProcessingStrategy[Any]:
...
+                except Exception:
+                    raise ValueError(f"Kafka topic {self.stream_name} has no associated schema")
+
+                now = time.time()
+                return RoutedValue(
+                    route=Route(source=self.source, waypoints=[]),
+                    payload=StreamsMessage(
+                        payload=value, headers=headers, timestamp=now, schema=schema
```
fpacifici, line 84: We should pick up the timestamp of the message coming from Kafka at the beginning. Not the consumer timestamp.

sentry_streams/sentry_streams/adapters/arroyo/msg_wrapper.py
```diff
@@ -0,0 +1,51 @@
...
+from sentry_streams.pipeline.message import Message as StreamsMessage
+
+TPayload = TypeVar("TPayload")
+
+
+class MessageWrapper(ProcessingStrategy[Union[FilteredPayload, TPayload]]):
+    """
+    Custom processing strategy which either produces an incoming message via a given Producer
+    if the Route of the message matches this strategy's Route, or forwards the message
+    to the next strategy provided.
```
fpacifici, line 17: This is the same docstring of the `Forwarder`  strategy above

sentry_streams/sentry_streams/adapters/arroyo/msg_wrapper.py
```diff
@@ -0,0 +1,51 @@
...
+        self.__route = route
+
+    def submit(self, message: Message[Union[FilteredPayload, TPayload]]) -> None:
+        now = time.time()
+        if not isinstance(message.payload, FilteredPayload):
+            msg = StreamsMessage(message.payload, [], now, None)
+
+            routed_msg: Message[RoutedValue] = Message(
+                Value(committable=message.value.committable, payload=RoutedValue(self.__route, msg))
+            )
```
fpacifici, line 35: This is a good sign that we will not stick with arroyo in the long run:
StreamsMessage is the payload of RoutedMessage 
which  is the payload of the value of Message. 
That's a lot

sentry_streams/sentry_streams/adapters/arroyo/reduce.py
```diff
@@ -90,7 +90,7 @@ def add(self, value: Any) -> Self:
                 self.offsets[partition] = max(offsets[partition], self.offsets[partition])
 
             else:
-                self.offsets.update(offsets)
+                self.offsets[partition] = offsets[partition]
```
fpacifici, line 93: ?

sentry_streams/sentry_streams/adapters/arroyo/consumer.py
ayirr7: yeah it was a typo, done

sentry_streams/sentry_streams/adapters/arroyo/msg_wrapper.py
ayirr7: Yeah unfortunately had to alias our top-level API's `Message` as a `StreamsMessage` in order to not clash with Arroyo's `Message`

ayirr7: I had kind of hoped for a solution where all of it can be managed at the top-level API code, but given that we want to expose _some_ metadata to the user, like processing time, the `StreamsMessage` type has to be managed in the Arroyo steps as well

sentry_streams/sentry_streams/adapters/arroyo/reduce.py
ayirr7: If you expand the code snippet, there was a bug, that got fixed here

sentry_streams/sentry_streams/pipeline/msg_parser.py
ayirr7: I'll come back to this after PTO in another PR. Thanks!

sentry_streams/sentry_streams/pipeline/chain.py
```diff
@@ -69,34 +74,37 @@ def build_step(self, name: str, ctx: Pipeline, previous: Step) -> Step:
...
 
     def build_step(self, name: str, ctx: Pipeline, previous: Step) -> Step:
         return FilterStep(name=name, ctx=ctx, inputs=[previous], function=self.function)
 
 
 @dataclass
-class FlatMap(Applier[TIn, TOut], Generic[TIn, TOut]):
-    function: Union[Callable[[TIn], TOut], str]
+class FlatMap(Applier[Message[TIn], Message[TOut]], Generic[TIn, TOut]):
+    function: Union[Callable[[Message[TIn]], Message[TOut]], str]
```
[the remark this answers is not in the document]
ayirr7: Gonna come back to this tmrw

sentry_streams/sentry_streams/adapters/arroyo/consumer.py
ayirr7: should be fixed now

sentry_streams/tests/adapters/arroyo/test_adapter.py
```diff
@@ -51,7 +53,12 @@ def test_kafka_sources() -> None:
     assert sources.get_consumer("source1") is not None
 
 
-def test_adapter(broker: LocalBroker[KafkaPayload], pipeline: Pipeline) -> None:
+def test_adapter(
+    broker: LocalBroker[KafkaPayload],
+    pipeline: Pipeline,
+    metric: IngestMetric,
```
ayirr7, line 59: I'll probably stop spamming the `IngestMetric` type everywhere in tests and examples, and use a different one instead

fpacifici (approved):

sentry_streams/sentry_streams/pipeline/chain.py
```diff
@@ -111,9 +119,37 @@ def build_step(self, name: str, ctx: Pipeline, previous: Step) -> Step:
...
+            ctx=ctx,
+            inputs=[previous],
+            function=msg_serializer,
+        )
+
+
 @dataclass
 class Batch(
-    Applier[InputType, MutableSequence[InputType]],
+    Applier[Message[InputType], Message[MutableSequence[InputType]]],
```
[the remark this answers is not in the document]
ayirr7: In order to append items to it

sentry_streams/sentry_streams/pipeline/chain.py
```diff
@@ -111,9 +119,37 @@ def build_step(self, name: str, ctx: Pipeline, previous: Step) -> Step:
         )
 
 
+@dataclass
+class Parser(Applier[Message[bytes], Message[TOut]], Generic[TOut]):
```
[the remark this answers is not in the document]
ayirr7: done

sentry_streams/sentry_streams/pipeline/chain.py
```diff
@@ -111,9 +119,37 @@ def build_step(self, name: str, ctx: Pipeline, previous: Step) -> Step:
...
+        return MapStep(
+            name=name,
+            ctx=ctx,
+            inputs=[previous],
+            function=msg_parser,
+        )
+
+
+@dataclass
+class Serializer(Applier[Message[TIn], bytes], Generic[TIn]):
```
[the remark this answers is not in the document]
ayirr7: done

PR #102, opened 2025-05-07
Add workflow to build the wheel containing the rust binary

fpacifici:
The previous PR introduced the rust adapter in a rust/python mixed project
with Maturin. 

It did not upate the release github workflow to produce a wheel that contained
both the rust binary and the python code.

This PR fixes the release:
- Uses the pyo3 action to build wheels for linux, macos on both x86_64 and arm
- Remove the auto-init option from pyo3 that is ok in tests but not recommended
  for release. This introduced the need to initialize the python interpreter 
  inside the tests manually.


.github/workflows/build-streams.yaml
```diff
@@ -7,25 +7,122 @@ on:
...
+          working-directory: ./sentry_streams
+          before-script-linux: |
+            cmake --version
+            # If we're running on rhel centos, install needed packages.
+            if command -v yum &> /dev/null; then
+                yum update -y && yum install -y openssl openssl-devel pkgconfig
+
+            else
+                # If we're running on debian-based system.
+                apt update -y && apt-get install -y libssl-dev openssl pkg-config
```
fpacifici, line 43: The container this action runs does not have openssl installed.

.github/workflows/build-streams.yaml
```diff
@@ -7,25 +7,122 @@ on:
...
+
+
+  macos:
+    permissions: read-all
+    timeout-minutes: 5
+    runs-on: ${{ matrix.platform.runner }}
+    strategy:
+      matrix:
+        platform:
+          - runner: macos-13
```
hubertdeng123, line 63: Is there any reason why we're using macos-13 for x86 and macos-latest for arm?

.github/workflows/build-streams.yaml
```diff
@@ -7,25 +7,122 @@ on:
...
+      - uses: actions/setup-python@v5
+        with:
+          python-version: 3.11
+      - name: Build wheels
+        uses: PyO3/maturin-action@v1
+        with:
+          target: ${{ matrix.platform.target }}
+          args: --release --out dist --find-interpreter
+          sccache: ${{ !startsWith(github.ref, 'refs/tags/') }}
+          manylinux: auto
```
hubertdeng123, line 33: I don't think we want to set `manylinux: auto` here, better to be specific and pin this since we're publishing a wheel here and don't know if we want to trust auto here

fpacifici: done. set to 2014

.github/workflows/build-streams.yaml
fpacifici: no, fixed.

.github/workflows/build-streams.yaml
```diff
@@ -7,25 +7,109 @@ on:
...
-
-        - name: "Prepare artifacts"
-          run: |
-            make build-streams
-
-        - uses: actions/upload-artifact@v4
-          with:
-                name: ${{ github.sha }}
-                path: sentry_streams/dist/*
+      - uses: actions/upload-artifact/merge@v4
```
hubertdeng123, line 110: nit: one last conversion from version to sha that was missed

hubertdeng123 (commented):
one small comment, otherwise looking good to me

.github/workflows/build-streams.yaml
fpacifici: done

hubertdeng123 (approved):

PR #104, opened 2025-05-12
Support routing in the Rust adapter

fpacifici:
Implement the router in the Rust adapter.

The routing function is expressed in Python in the same format
as the one used in the Python adapter.
The rust adapter calls the python routing function and then appends
the new waypoint to the Route object.

The RoutedValue object is never visible to the Python code. This is
intentional as we can keep all the routing logic inside the Rust 
adapter.


sentry_streams/src/callers.rs
```diff
@@ -0,0 +1,50 @@
+use crate::routes::RoutedValue;
+use pyo3::prelude::*;
+use sentry_arroyo::types::Message;
+
+/// Executes a Python callable with an Arroyo message containing Any and
+/// returns the result.
+pub fn call_python_function(
```
evanh, line 7: Is this meant to be a generic wrapper around python calls? If so, maybe this should be in a mod call "python"?

sentry_streams/src/routers.rs
```diff
@@ -0,0 +1,106 @@
...
+fn route_message(
+    route: &Route,
+    callable: &Py<PyAny>,
+    message: Message<RoutedValue>,
+) -> Result<Message<RoutedValue>, SubmitError<RoutedValue>> {
+    if message.payload().route != *route {
+        return Ok(message);
+    }
+    let dest_route = call_python_function(callable, &message);
+    Python::with_gil(|py| match dest_route {
```
evanh, line 17: Is it possible to move the Python call out of this main loop? This is more of a stylistic thing:

```
match dest_route {
    Ok(dest_route) => {
        let new_waypoint = extra_python_string(dest_route);
        ...
}
```

fpacifici: You are right, it should work.

sentry_streams/src/callers.rs
fpacifici: I meant this more as a place to keep the different ways to call back into python function, so not everything python. Like python exposed data structures are in the modules where they belong to like `routes.rs` rather than in a general `python` module.

sentry_streams/src/routers.rs
fpacifici: done

sentry_streams/src/routers.rs
```diff
@@ -0,0 +1,106 @@
...
+    message: Message<RoutedValue>,
+) -> Result<Message<RoutedValue>, SubmitError<RoutedValue>> {
+    if message.payload().route != *route {
+        return Ok(message);
+    }
+    let dest_route = call_python_function(callable, &message);
+    match dest_route {
+        Ok(dest_route) => Python::with_gil(|py| {
+            let new_waypoint = dest_route.extract::<String>(py).unwrap();
+            message.try_map(|payload| Ok(payload.add_waypoint(new_waypoint.clone())))
```
untitaker, line 20: from what i can tell this line can run outside of python GIL. we may not have other python threads running as part of the consumer but holding the GIL too long will impact the performance of the python SDK's background threads if nothing else

untitaker (approved):

sentry_streams/src/routers.rs
fpacifici: you are right. Will change

fpacifici: done

PR #105, opened 2025-05-15
feat: Basic filter step for the Rust adapter

ayirr7:
- Adds a Filter step in the Rust Arroyo adapter so that Filter steps defined in a Python-defined pipeline can be run by Rust Arroyo
- Adds a new Filter processing strategy in streams to do so
- Tests and slightly updated example


sentry_streams/src/filter_step.rs
```diff
@@ -0,0 +1,66 @@
...
+use pyo3::{Py, PyAny, Python};
+use sentry_arroyo::processing::strategies::{
+    CommitRequest, ProcessingStrategy, StrategyError, SubmitError,
+};
+use sentry_arroyo::types::Message;
+use std::time::Duration;
+
+pub struct Filter<N> {
+    pub callable: Py<PyAny>,
+    pub next_step: N,
```
fpacifici, line 11: I think this can be more specific: `Box<dyn ProcessingStrategy<RoutedValue>>` 

The next step has to be a ProcessingStrategy and we know we only pass around RoutedValue

fpacifici (commented):
Goes in the right direction. Thanks.

Please do not forget the python interface 
https://github.com/getsentry/streams/blob/main/sentry_streams/sentry_streams/rust_streams.pyi#L39-L47

Also please add some unit tests.

sentry_streams/src/callers.rs
```diff
@@ -8,7 +8,7 @@ pub fn call_python_function(
     callable: &Py<PyAny>,
     message: &Message<RoutedValue>,
 ) -> Result<Py<PyAny>, PyErr> {
-    Python::with_gil(|py| {
+    Python::with_gil(|py: Python<'_>| {
```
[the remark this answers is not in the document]
ayirr7: no problem, just redundant type hint added by accident

sentry_streams/src/filter_step.rs
ayirr7: What's the reasoning for the `dyn` here?

fpacifici: ProcessingStrategy is a trait, I think dyn is required.

sentry_streams/sentry_streams/rust_streams.pyi
```diff
@@ -45,6 +45,8 @@ class RuntimeOperator:
     ) -> Self: ...
     @classmethod
     def Router(cls, route: Route, function: Callable[[Any], str]) -> Self: ...
+    @classmethod
+    def Filter(cls, route: Route, function: Callable[[Any], bool]) -> Self: ...
```
fpacifici, line 49: Can we restrict these types to `Message` now that you introduced Message ?

sentry_streams/src/filter_step.rs
```diff
@@ -0,0 +1,70 @@
...
+pub struct Filter {
+    pub callable: Py<PyAny>,
+    pub next_step: Box<dyn ProcessingStrategy<RoutedValue>>,
+    pub route: Route,
+}
+
+impl Filter {
+    /// A strategy that takes a callable, and applies it to messages
+    /// to either filter the message out or submit it to the next step.
+    /// The callable is expected to take a Message<Py<PyAny>>
```
fpacifici, line 18: It takes `Message<RoutedValue>` not `Message<Py<PyAny>>`

sentry_streams/src/filter_step.rs
```diff
@@ -0,0 +1,70 @@
...
+            self.next_step.submit(message)
+        } else {
+            let filtered = Python::with_gil(|py: Python<'_>| -> bool {
+                let python_payload = message.payload().payload.clone_ref(py);
+                let py_res = self.callable.call1(py, (python_payload,));
+
+                // TODO: Create an exception for Invalid messages in Python
+                // This now crashes if the Python code fails.
+                let boolean = py_res.unwrap();
+                boolean.is_truthy(py).unwrap()
```
fpacifici, line 51: You can do this https://github.com/getsentry/streams/blob/main/sentry_streams/src/routers.rs#L22-L29

sentry_streams/src/transformer.rs
```diff
@@ -100,4 +117,70 @@ mod tests {
             assert_messages_match(py, expected_messages, actual_messages.deref());
         });
     }
+
+    #[test]
```
fpacifici, line 121: I'd put this in the filter_step module as rust tests go together with the code they test.

fpacifici (approved):

sentry_streams/src/filter_step.rs
ayirr7: Right, I wasn't sure if we were referring to the user-defined function (which is currently in Python) or the `callable` which only assumes a single argument of `Py<PyAny>`

sentry_streams/src/filter_step.rs
ayirr7: I think we can't do this because the `message` gets moved, and then we need to do `next_step.submit(message)`

PR #106, opened 2025-05-16
(not ready for review) Protobuf types

ayirr7:


.pre-commit-config.yaml
```diff
@@ -43,6 +43,8 @@ repos:
           types-pyYAML,
           types-jsonschema,
           "sentry-kafka-schemas>=1.2.0",
+          "types-protobuf",
+          "sentry-protos>=0.1.74"
```
fpacifici, line 47: This is FSL, we cannot import it in an Apache 2.0 library

sentry_streams/sentry_streams/pipeline/msg_parser.py
```diff
@@ -21,7 +23,13 @@ def msg_parser(msg: Message[bytes]) -> Any:
...
     payload = msg.payload
 
-    return json.dumps(payload).encode("utf-8")
+    if schema_type is MessageSchema.PROTOBUF:
+        assert isinstance(payload, ProtoMessage)
+        return payload.SerializeToString()
+    elif schema_type is MessageSchema.JSON:
+        return json.dumps(payload).encode("utf-8")
+    else:
+        raise Exception(f"Unknown codec / message schema type {schema_type}")
```
fpacifici, line 35: I think you can catch this with the type checker with `assert_never`

fpacifici (commented):
I like the design. As long as we have an explicit Parser step, it makes sense to have a Serializer. We can revisit whether to expose those steps later.

Please consider not depending on sentry-protos. It seems you only need them for examples. You should be able to create some example protobuf messages without.

.pre-commit-config.yaml
untitaker: Do I understand it right that this is only used for tests? in that case why not write your own protobuf schema?

ayirr7: I can do that, I just relied on existing ones out of ease of use

PR #107, opened 2025-05-18
Add a Python Delegate strategy to implement Streaming steps in python

fpacifici:
As we are moving to the rust runtime, there are still several python steps that 
need to be ported and would take time to port.
Examples are the Reduce step and a multiprocess step (which is not present in Arroyo
Rust).

This simplifies this process by allowing us to implement Streaming steps in Python
and wiring them up into a Rust powered Arroyo strategy.

This follows the same design we have in Snuba for Rust snuba consumers with python
logic that relies on python multi processing:
https://github.com/getsentry/snuba/blob/master/rust_snuba/src/strategies/python.rs

This PR includes:
- A PYthon delegate interface `RustOperatorDelegate` which defines the interface
for the python logic that is executed by the Rust adapter.
- A Rust Arroyo strategy that serves as an adapter for the Python logic. This one
  implements the ProcessingStrategy trait and is able to send calls to the 
  `RustOperatorDelegate`. 

This is meant to simplify porting of the steps we have and should be replaced by a
proper native rust implementation for all steps.


sentry_streams/src/fake_strategy.rs
```diff
@@ -1,33 +1,78 @@
...
 };
-use sentry_arroyo::types::Message;
+use sentry_arroyo::types::{AnyMessage, BrokerMessage, InnerMessage, Message, Partition, Topic};
+use std::collections::HashMap;
 use std::sync::{Arc, Mutex};
 use std::time::Duration;
 
 pub struct FakeStrategy {
     pub submitted: Arc<Mutex<Vec<Py<PyAny>>>>,
+    pub reject_message: bool,
```
ayirr7, line 15: These look to be separate changes from the `RustOperatorDelegate`-related work? What's the purpose of these changes being included in this PR?

ayirr7:
Some thoughts/qs:

- We don't have any examples of extending `RustOperatorDelegate`, right? Should we add one? For example, I haven't quite been able to understand how we expect to return the processing results in the `poll()` when our Python processing strategy does not return anything in `poll()`. I'm probably misunderstanding something. 
- Haven't really wrapped my head around what the `submit` should look like for a class which extends delegate. What is its purpose?

fpacifici:
Ok will add an example

sentry_streams/sentry_streams/adapters/arroyo/rust_step.py
```diff
@@ -0,0 +1,89 @@
...
+    following step to send message to would have been considerably more
+    complex.
+    """
+
+    @abstractmethod
+    def submit(self, message: Message[TIn], committable: Committable) -> None:
+        """
+        Send a message to this step for processing.
+
+        This method accumulate the message with work to be done.
```
ayirr7, line 55: `This method accumulate the message with work to be done.` I think this could be a little clearer. If my question in the separate comment re: state is on the correct lines maybe we describe how `submit` is an API that takes in a message, and the delegate is responsible for accumulating it or adding it to some state it is tracking.

sentry_streams/src/python_operator.rs
```diff
@@ -0,0 +1,520 @@
...
+            }
+            Err(e) => Err(StrategyError::Other(Box::new(e))),
+        }
+    }
+
+    fn terminate(&mut self) {
+        self.next_strategy.terminate();
+    }
+
+    fn join(&mut self, timeout: Option<Duration>) -> Result<Option<CommitRequest>, StrategyError> {
```
ayirr7, line 224: Simple docstring here could be good to have

this step
- triggers flush (with a timeout) on the delegate, thus returning any final results and clearing any state and closing resources
- submits those results to the next step

ayirr7: though, is there a reason why it doesn't call `join` on the next step?

sentry_streams/sentry_streams/adapters/arroyo/rust_step.py
```diff
@@ -0,0 +1,89 @@
...
+    poll or join. The Rust adapter takes care to send them to
+    the next step in the pipeline.
+
+    Following the arroyo interface and providing this class with the
+    following step to send message to would have been considerably more
+    complex.
+    """
+
+    @abstractmethod
+    def submit(self, message: Message[TIn], committable: Committable) -> None:
```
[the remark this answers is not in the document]
fpacifici: It is not intentional. 
We have two types of usages of the adapter: sync and async. The synchronous one immediately produces the result when you submit. It is like the filter that calls next.submit from the submit method. The async one is the one that accumulate messages and perform work in poll.

Both can support these cardinalities: 1:1, 1:n, 1:0, n:1.

I was considering creating multiple interfaces for those cases but they turned out to be a lot for marginal gain as they introduced a ton of special cases in the rust code, mostly because of the way it processes errors (submit can only return SubmitError (which can only be a MessageRejected or InvalidMessage). This did not fit well with the 1:n cases. 
So in the end I found that a simpler interface that supports all cases above was easier both to write and to use.

sentry_streams/sentry_streams/adapters/arroyo/rust_step.py
fpacifici: Will try to rephrase.

sentry_streams/src/python_operator.rs
```diff
@@ -0,0 +1,520 @@
...
+            for (partition, offset) in message.committable() {
+                committable.insert(partition, offset);
+            }
+
+            Python::with_gil(|py| {
+                let payload = message.payload().payload.clone_ref(py);
+                let py_committable = convert_committable_to_py(py, committable).unwrap();
+                match self
+                    .processing_step
+                    .call_method1(py, "submit", (payload, py_committable))
```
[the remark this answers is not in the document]
fpacifici: Yes.
The contract is that we call Submit on the delegate for each message and periodically we call poll.
It is up to the delegate to decide how to manage and process messages.

sentry_streams/src/python_operator.rs
fpacifici: > though, is there a reason why it doesn't call join on the next step?\

I think it is a bug. Thanks

sentry_streams/sentry_streams/adapters/arroyo/rust_step.py
fpacifici: rephrased

sentry_streams/src/python_operator.rs
fpacifici: fixed

fpacifici:
@ayirr7 , I added an example implementation of `RustOperatorDelegate` and rephrased the docstring. Please let me know if this is clearer.

ayirr7 (approved):

PR #113, opened 2025-05-22
ref(ci): Disable extension-module by default

untitaker:
Since people interact with cargo directly in local development, i'd argue for the sake of ergonomics that the features should default to what is needed in local development. For the release job we can have a longer list of arguments. Going to run the build-streams job as well in CI, because it'd suck if any breakage is only revealed during release.

### DIFF
--- a/.github/workflows/build-streams.yaml
+++ b/.github/workflows/build-streams.yaml
@@ -1,6 +1,7 @@
 name: build sentry_streams
 
 on:
+  pull_request:
   push:
     branches:
       - main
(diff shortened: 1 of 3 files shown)


.github/workflows/build-streams.yaml
```diff
@@ -1,6 +1,7 @@
 name: build sentry_streams
 
 on:
+  pull_request:
```
fpacifici, line 4: Is this intentional or is it a left over ?

sentry_streams/Cargo.toml
```diff
@@ -19,7 +19,6 @@ crate-type = ["cdylib"]
 
 [features]
 extension-module = ["pyo3/extension-module"]
-default = ["extension-module"]
```
fpacifici, line 22: Did you try to release without this ?

.github/workflows/build-streams.yaml
untitaker: this is intentional to test the release build within this PR and show that it works.

fpacifici (approved):

PR #116, opened 2025-05-25
Move the Message data structure to Rust

fpacifici:
This is the foundation of the work to manage Message structures in Rust.
There is a more refactoring  work to do to have a proper Rust Message abstraciton
though this is step one we can build on top of.

We are trying to move more streaming processing logic to Rust (for efficiency).
At the same time, we want to build Python streaming application,
Rust streaming application and hybrid streaming application where some
steps are in Python and some are in Rust.

The requirement above imply both Rust and Python must understand the Message
data structure, thus this has to be defined in Rust and exposed via Pyo3.
Rust orchestrate the pipeline. As such it will create the message instance
after the consumer. 

Having both Rust and Python streaming steps means we need multiple types of
messages (we cannot use generics as we cannot expose to python a rust generic
structure). Different message types will be able to be processed by some 
operators and not by others. 

Having oth rust code and python code processing the same message means the
message msut be able to move in and out of the Python memory otherwise 
Rust has to take the Gil every time it needs to read the content of the message.

This PR starts addressing some of the problems above:
- Moves the Message class from PYthon to Rust and introduces two message
  types: PyAny and Bytes Array. This allows us to change all the Rust
  steps interface to manage a message they can understand rather than an
  opaque PyAny.
- Provides some transformation function to copy messages in and out of
  Python memory so the same message can be handed over to Python and back
  thus being processable through hybrid pipelines. (source and sink being 
  Rust means all pipelines are hybrid already). This opens the door to
  more Rust native steps. Transformation requires copy so it is not the
  most effiient way. Though there are ways to optimize this going on.
- Introduces a wrapper in Python to give Python code the illusion that the
  Message is generic with respect to the payload type. This cannot be done
  with a Rust generic.
- Replaces the usage of the old Message class in Python code and adapts the
   usage of the schema.

Next steps on top of this (see the TODOs in `messages.rs`).
- Remove copy paste. Create the hierarchy of messages as an Enum in Rust
- Introduce a unified transformation layer between Python and Rust so
  we can optimize translation by reducing copy.
- Create two variants of the same message to make it available to both
  Rust and Python.

### DIFF
--- a/sentry_streams/sentry_streams/pipeline/message.py
+++ b/sentry_streams/sentry_streams/pipeline/message.py
@@ -44,15 +44,6 @@ class Message(ABC, Generic[TPayload]):
             and self.schema == other.schema
         )
 
-    def __repr__(self) -> str:
-        return (
-            f"{self.__class__.__name__}("
-            f"payload={self.payload!r}, "
-            f"headers={self.headers!r}, "
-            f"timestamp={self.timestamp!r}, "
-            f"schema={self.schema!r})"
-        )
-
 
 class PyMessage(Generic[TPayload], Message[TPayload]):
     """
--- a/sentry_streams/sentry_streams/pipeline/message.py
+++ b/sentry_streams/sentry_streams/pipeline/message.py
@@ -1,21 +1,60 @@
 from __future__ import annotations
 
+from abc import ABC, abstractmethod
 from copy import deepcopy
-from typing import (
-    Generic,
-    Optional,
-    Sequence,
-    Tuple,
-    TypeVar,
-    cast,
-)
+from typing import Generic, Optional, Sequence, Tuple, TypeVar, cast
 
-from sentry_streams.rust_streams import PyAnyMessage
+from sentry_streams.rust_streams import PyAnyMessage, RawMessage
 
 TPayload = TypeVar("TPayload")
 
 
-class Message(Generic[TPayload]):
+class Message(ABC, Generic[TPayload]):
+    @property
+    @abstractmethod
+    def payload(self) -> TPayload:
+        raise NotImplementedError
+
+    @property
+    @abstractmethod
+    def headers(self) -> Sequence[Tuple[str, bytes]]:
+        raise NotImplementedError
+
+    @property
+    @abstractmethod
+    def timestamp(self) -> float:
+        raise NotImplementedError
+
+    @property
+    @abstractmethod
+    def schema(self) -> str | None:
+        raise NotImplementedError
+
+    @abstractmethod
+    def deepcopy(self) -> Message[TPayload]:
+        raise NotImplementedError
+
+    def __eq__(self, other: object) -> bool:
+        if not isinstance(other, Message):
+            return False
+        return (
+            self.payload == other.payload
+            and self.headers == other.headers
+            and self.timestamp == other.timestamp
+            and self.schema == other.schema
+        )
+
+    def __repr__(self) -> str:
+        return (
+            f"{self.__class__.__name__}("
+            f"payload={self.payload!r}, "
+            f"headers={self.headers!r}, "
+            f"timestamp={self.timestamp!r}, "
+            f"schema={self.schema!r})"
+        )
+
+
+class PyMessage(Generic[TPayload], Message[TPayload]):
     """
     A wrapper for the Rust PyAnyMessage to make the payload generic.
     We cannot export classes via PyO3 that are generic in Python.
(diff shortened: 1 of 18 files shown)


untitaker (approved):

sentry_streams/src/messages.rs
```diff
@@ -0,0 +1,434 @@
...
+
+    #[pyo3(get, set)]
+    pub timestamp: f64,
+
+    #[pyo3(get, set)]
+    pub schema: Option<String>,
+}
+
+#[pymethods]
+impl RawMessage {
```
evanh, line 174: All of the functions here seem to be for creating the RawMessage from Python objects, and converting the message into Python objects. Can we use this interface to go from Rust to Rust?

I might be misunderstanding how the Rust API handles messages.

evanh: What I mean is, all the functions return Python object (the `headers` call for example). If this is always converted to a Python object, how does the Rust to Rust functionality work?

evanh: Yes OK that makes sense.

sentry_streams/sentry_streams/pipeline/msg_parser.py
```diff
@@ -1,21 +1,31 @@
...
+from sentry_kafka_schemas import get_codec
+from sentry_kafka_schemas.codecs import Codec
+
+from sentry_streams.pipeline.message import Message, PyRawMessage
 
 # TODO: Push the following to docs
 # Standard message decoders and encoders live here
 # These are used in the defintions of Parser() and Serializer() steps, see chain/
 
+CODECS: MutableMapping[str, Codec[Any]] = {}
```
ayirr7, line 13: Is the idea here that we're going to build our own Codecs?

sentry_streams/sentry_streams/pipeline/message.py
```diff
@@ -1,27 +1,140 @@
...
+
+
+class Message(ABC, Generic[TPayload]):
+    """
+    A generic class to represent multiple types of messages in the pipeline.
+    Streaming Steps should access the message via this class.
+
+    The actual Message classes are defined in Rust and are exported to Python
+    via pyo3. This class and its subclasses wrap the Rust classes so that we
+    can make the payload time Generic. It is not possible to create, via pyo3,
```
ayirr7, line 19: typo; 'time'?

sentry_streams/sentry_streams/pipeline/message.py
```diff
@@ -1,27 +1,140 @@
...
+            self.payload == other.payload
+            and self.headers == other.headers
+            and self.timestamp == other.timestamp
+            and self.schema == other.schema
+        )
+
+
+class PyMessage(Generic[TPayload], Message[TPayload]):
+    """
+    A wrapper for the Rust PyAnyMessage to make the payload generic.
```
ayirr7, line 67: Does that mean we'll have types like `PyMessage[IngestMetric]`? I guess I still don't fully understand how adding this wrapper class helps us achieve the generic

sentry_streams/sentry_streams/pipeline/message.py
```diff
@@ -1,27 +1,140 @@
...
+            deepcopy(self.inner.payload),
+            deepcopy(self.inner.headers),
+            self.inner.timestamp,
+            self.inner.schema,
+        )
+
+
+class PyRawMessage(Message[bytes]):
+    """
+    A wrapper for the Rust RawMessage so `RawMessage` extends `Messasge`.
```
ayirr7, line 106: Is PyRawMessage the container for bytes in user/API code in Python and RustRawMessage the container for bytes in user/API code in Rust? A little confused about the two's relationship

sentry_streams/src/messages.rs
```diff
@@ -0,0 +1,434 @@
...
+    let seq = list.into_sequence();
+    Ok(seq.unbind())
+}
+
+/// Represents a message with any Python object as payload. This message type
+/// is meant to be processed by Python operators. Rust operators should consider
+/// The payload as opaque and should not transform it.
+///
+/// The message can be created both by Python code or Rust code. When Rust code
+/// instantiates it, it has to convert it into a Py<PyAnyMessage> before handing
```
ayirr7, line 84: Am I understanding correctly that Rust would hand a Message to Python code when, for example, running a user-defined Python function on a Message that is the output from a previous Rust strategy step?

sentry_streams/sentry_streams/pipeline/msg_parser.py
fpacifici: This is mostly to take the codec out of the message to avoid creating the codec each time we need to use and reuse the instances. 

But yes we will need an abstract codec that can wrap the sentry-kafka-shemas ones.

sentry_streams/src/messages.rs
fpacifici: > Am I understanding correctly that Rust would hand a Message to Python code when, for example, running a user-defined Python function on a Message that is the output from a previous Rust strategy step?

Rust will hand a message to python code every time a user-defined Python function needs to process the message or when a platform defined python processing function (the Delegate) needs to process the message.

Regarding the previous step, that can be a Rust step or a Python step. 
In case the previous step is a rust step the payload type needs to be translatable to python. Otherwise python can only receive bytes it cannot make sense of. 

So we will have these combinations:
Messages:
1. Python only messages (they are opaque to rust)
2. Rust only message - contain rust data structures (they are opaque to python)
3. messages that can be manage by both via copy (for example bytes)
4.  messages that have the same memory structure in both and could be managed by both without copy (this requires some unsafe code and FFI usage on python but it is theoretically possible).

Steps:
- rust native. Supports 2,3,4
- rust steps with python custom logic/ Supports 1,3,4
- Python delegates. Supports 1,3,4

PR #120, opened 2025-05-27
Generalize ReduceDelegate

fpacifici:
Based on https://github.com/getsentry/streams/pull/119

While buildign a RustOperatorDelegate for the RunTaskInMultiProcess
I noticed that the delegating logic was very similar to the Reduce
step so I make the `OutputRetriever` and the `ReduceDelegate` reusable.

This PR:
- replaces the OutputRetriever and ReduceDelegate custom logic with
  functions passed to the two classes
- Moves the generic version of the two into rust_step.py
- Fixes the return type iof the RustOperatorDeelgate. We were returning
  Message, but that is a type Rust cannot understand.


sentry_streams/sentry_streams/adapters/arroyo/rust_step.py
```diff
@@ -161,14 +161,132 @@ def __prepare_output(self) -> Iterable[Tuple[Message[TOut], Committable]]:
...
+    """
+    This delegate wraps an existing Python Arroyo strategy so that it can be
+    used as it is in a Rust consumer.
+    Objects of this class are provided a strategy to wrap, a Callable to
+    transform input messages if a transformation is needed and an
+    `OutputRetriever` to collect the results from the inner strategy and
+    make them available to rust.
+
+    The message flow looks like this:
+    1. The Rust Arroyo Strategy receives a streaming message to process.
```
ayirr7, line 247: Thanks for outlining the steps here. It is much clearer now

ayirr7 (commented):
This overall looks good to me. I'll wait for any others' who might have feedback before I approve

ayirr7 (approved):

PR #126, opened 2025-06-03
feat(GCS-sink): Use RunTaskInThreads to write to GCS

ayirr7:
Switches from single threaded write to GCS to a RunTaskInThreads GCS sink. Also adds some tracing around who's using the GIL. 

Seems to fix an issue where GCS writer could not acquire GIL

### DIFF
--- a/sentry_streams/Cargo.toml
+++ b/sentry_streams/Cargo.toml
@@ -16,6 +16,7 @@ rdkafka = { version = "0.37.0", features = ["cmake-build", "tracing"] }
 anyhow = "1.0.98"
 reqwest = "0.12.15"
 tokio = "1.45.0"
+log = "0.4.27"
 
 [lib]
 name = "rust_streams"
--- a/sentry_streams/src/callers.rs
+++ b/sentry_streams/src/callers.rs
@@ -15,20 +15,12 @@ pub fn call_python_function(
     Ok(
         traced_with_gil("call python fn", |py| match message.payload().payload {
             PyStreamingMessage::PyAnyMessage { ref content } => {
-                let start = Instant::now();
                 let res = callable.call1(py, (content.clone_ref(py),));
-                let finish = Instant::now();
-
-                println!("EXECUTINO TIME CALL PYTHON FN {:?}", finish - start);
                 res
             }
 
             PyStreamingMessage::RawMessage { ref content } => {
-                let start = Instant::now();
                 let res = callable.call1(py, (content.clone_ref(py),));
-                let finish = Instant::now();
-
-                println!("EXECUTINO TIME CALL PYTHON FN {:?}", finish - start);
                 res
             }
         })?
--- a/sentry_streams/src/callers.rs
+++ b/sentry_streams/src/callers.rs
@@ -1,5 +1,3 @@
-use std::time::Instant;
-
 use crate::helper::traced_with_gil;
 use crate::messages::PyStreamingMessage;
 use crate::routes::RoutedValue;
--- a/sentry_streams/src/callers.rs
+++ b/sentry_streams/src/callers.rs
@@ -1,4 +1,4 @@
-use crate::helper::traced_with_gil;
+use crate::utils::traced_with_gil;
 use crate::messages::PyStreamingMessage;
 use crate::routes::RoutedValue;
 use pyo3::prelude::*;
@@ -10,20 +10,18 @@ pub fn call_python_function(
     callable: &Py<PyAny>,
     message: &Message<RoutedValue>,
 ) -> Result<PyStreamingMessage, PyErr> {
-    Ok(
-        traced_with_gil("call python fn", |py| match message.payload().payload {
+    Ok(traced_with_gil("call_python_function", |py| {
+        match message.payload().payload {
             PyStreamingMessage::PyAnyMessage { ref content } => {
-                let res = callable.call1(py, (content.clone_ref(py),));
-                res
+                callable.call1(py, (content.clone_ref(py),))
             }
 
             PyStreamingMessage::RawMessage { ref content } => {
-                let res = callable.call1(py, (content.clone_ref(py),));
-                res
+                callable.call1(py, (content.clone_ref(py),))
             }
-        })?
-        .into(),
-    )
+        }
+    })?
+    .into())
 }
 
 /// Executes a Python callable with an Arroyo message containing Any and
@@ -32,12 +30,14 @@ pub fn call_any_python_function(
     callable: &Py<PyAny>,
     message: &Message<RoutedValue>,
 ) -> Result<Py<PyAny>, PyErr> {
-    traced_with_gil("call any python fn", |py| match message.payload().payload {
-        PyStreamingMessage::PyAnyMessage { ref content } => {
-            callable.call1(py, (content.clone_ref(py),))
-        }
-        PyStreamingMessage::RawMessage { ref content } => {
-            callable.call1(py, (content.clone_ref(py),))
+    traced_with_gil("call_any_python_function", |py| {
+        match message.payload().payload {
+            PyStreamingMessage::PyAnyMessage { ref content } => {
+                callable.call1(py, (content.clone_ref(py),))
+            }
+            PyStreamingMessage::RawMessage { ref content } => {
+                callable.call1(py, (content.clone_ref(py),))
+            }
         }
     })
 }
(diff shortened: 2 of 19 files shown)


sentry_streams/src/gcs_writer.rs
```diff
@@ -55,35 +59,45 @@ impl GCSWriter {
...
+        let route = message.payload().route.clone();
+        let actual_route = self.route.clone();
+
+        let bytes = traced_with_gil("writing to gcs", |py| pybytes_to_bytes(&message, py)).unwrap();
 
-        let res = client.post(&url).body(bytes).send();
+        Box::pin(async move {
+            if route != actual_route {
+                return Ok(message);
+            }
```
fpacifici, line 78: I think this is quite inefficient when we need to forward the message to the next step. 
Let's say we have two routes: one send small messages somewhere and one sends batches to GCS.
This way we would use threads to send the small messages to the next step. 

I think we should customize the runTaskInThreads to avoid sending messages for other routes to the threadpool.

Though please file a task for this, no need to do it here

sentry_streams/src/gcs_writer.rs
```diff
@@ -55,35 +59,45 @@ impl GCSWriter {
...
 
-        let response = res.unwrap();
-        let status = response.status();
+            let res: Result<reqwest::Response, reqwest::Error> =
+                client.post(&url).body(bytes).send().await;
```
fpacifici, line 81: How was this working before without await ?

sentry_streams/src/helper.rs
```diff
@@ -0,0 +1,26 @@
+use std::time::Instant;
+
+use pyo3::Python;
+
+pub fn traced_with_gil<F, R>(label: &str, function: F) -> R
```
fpacifici, line 5: I think having the ability to trace the usage of the GIL is really good and we should keep this ability to trace, but it has to be disabled by default and we need a way to enable it.

I'd recommend having a shared readonly config object we can pass around to keep the config values this function will use to decide whether to trace the gil usage or not.

fpacifici (commented):
Please see the comment on the tracing configuration.
We should keep the ability to instrument the GIL but we should make it configurable.

sentry_streams/src/gcs_writer.rs
ayirr7: Since we were using `RunTask` which requires a synchronous task to be run, I had to use the crate `blocking reqwest` and used its respective `Client`, `ClientBuilder` 

Now I am using the async `reqwest` crate

sentry_streams/src/helper.rs
fpacifici: alright

ayirr7: How do we feel, instead of having each function opt-into the `traced_with_gil`, we make it so that we only emit a warning when acquiring the GIL took over a certain duration? 

That way, we require everyone to go through the `traced_with_gil` but we selectively log

sentry_streams/src/callers.rs
```diff
@@ -9,15 +10,20 @@ pub fn call_python_function(
...
-            callable.call1(py, (content.clone_ref(py),))
-        }
-    })?
-    .into())
+    Ok(
+        traced_with_gil("call python fn", |py| match message.payload().payload {
+            PyStreamingMessage::PyAnyMessage { ref content } => {
+                let res = callable.call1(py, (content.clone_ref(py),));
+                res
+            }
```
fpacifici, line 18:
```
PyStreamingMessage::PyAnyMessage { ref content } => {
             callable.call1(py, (content.clone_ref(py),))
}
```

sentry_streams/src/helper.rs
fpacifici: ok

sentry_streams/src/messages.rs
```diff
@@ -251,7 +253,7 @@ pub enum PyStreamingMessage {
 
 impl Into<PyStreamingMessage> for Py<PyAny> {
     fn into(self) -> PyStreamingMessage {
-        Python::with_gil(|py| {
+        traced_with_gil("into pystreaming", |py| {
```
fpacifici, line 256: I would use the trace_with_gil function in unit tests as  well.

sentry_streams/src/gcs_writer.rs
```diff
@@ -55,35 +59,47 @@ impl GCSWriter {
...
+
+        let bytes = traced_with_gil("writing to gcs", |py| pybytes_to_bytes(&message, py)).unwrap();
 
-        let res = client.post(&url).body(bytes).send();
+        Box::pin(async move {
+            // TODO: This route-based forwarding does not need to be
+            // run with multiple threads. Look into removing this from the async task.
+            if route != actual_route {
+                return Ok(message);
+            }
```
fpacifici, line 80: Actually it would be trivial to fix it here. 
As you reintroduced the outer GCSSink strategy you can put this logic in the `submit` method there avoiding to pass it to inner.

fpacifici: Or is the issue due to the fact you cannot have two owners of the next step?

ayirr7: This

sentry_streams/src/filter_step.rs
```diff
@@ -57,24 +58,30 @@ impl ProcessingStrategy<RoutedValue> for Filter {
...
+                        Err(SubmitError::<RoutedValue>::InvalidMessage(InvalidMessage {
+                            partition: inner.partition,
+                            offset: inner.offset,
+                        }))
+                    }
+                    InnerMessage::AnyMessage(inner) => {
+                        panic!("Unexpected message type: {:?}", inner)
+                    }
+                },
+            }
```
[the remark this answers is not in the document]
ayirr7: I wanted to hold the GIL for even shorter time in the filter step

sentry_streams/src/operators.rs
```diff
@@ -104,20 +105,28 @@ pub fn build(
             route,
             bucket,
             object_file,
-        } => build_gcs_sink(route, next, &bucket, &object_file),
+        } => Box::new(GCSSink::new(
+            route.clone(),
+            next,
+            concurrency_config,
```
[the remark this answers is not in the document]
ayirr7: Yes, right now it's just sharing the config defined in operators.rs

sentry_streams/Cargo.toml
```diff
@@ -16,6 +16,7 @@ rdkafka = { version = "0.37.0", features = ["cmake-build", "tracing"] }
 anyhow = "1.0.98"
 reqwest = "0.12.15"
 tokio = "1.45.0"
+log = "0.4.27"
```
untitaker, line 19: only `tracing` should be necessary -- it exposes the same API `tracing::warn` instead of `log::warn` and so on

sentry_streams/src/filter_step.rs
```diff
@@ -57,24 +58,30 @@ impl ProcessingStrategy<RoutedValue> for Filter {
...
-                        }
-                    },
+                    Err(e) => return Err(e),
                 }
-            })
+            });
+            match res {
+                Ok(bool) => {
+                    if bool {
+                        let _ = self.next_step.submit(message);
```
untitaker, line 69: totally unrelated but it seems to me that backpressure is not handled correctly here
```suggestion
                        self.next_step.submit(message)?;
```

sentry_streams/src/gcs_writer.rs
ayirr7: @untitaker Did you all ever run into this in Rust Arroyo? Do yo know of any workarounds? 

Basically I want to be able to send messages on a different route to the next step (after `RunTaskInThreads`). The issue is that `RunTaskInThreads` step owns next_step

sentry_streams/Cargo.toml
ayirr7: sounds good

sentry_streams/src/gcs_writer.rs
untitaker: you can add an accessor on `RunTaskInThreads` such as:

```
fn next_step_mut(&mut self) -> &mut N { &mut self.next_step }
```

then you can use `self.inner.next_step_mut().submit(...)` in your GCS Sink

I don't think this usecase of poking into the next step has come up before, but i did find it occasionally annoying for observability that one step completely hides the next one. might be worth adding to the strategy interface.

sentry_streams/src/filter_step.rs
ayirr7: fixed, thanks

sentry_streams/src/messages.rs
ayirr7: done

PR #129, opened 2025-06-04
expose batch_size in config

victoria-yining-huang:
We want to support `batch_size` being passed in either from the application code, when a `Batch` class is instantiated (`Batch(batch_size=n)`), or when it's being passed in from the deployment_config file. 

There is an overriding logic:
- the value from the application code is the default value
- the value from config file takes higher precedence. This can represent values in different regions

### DIFF
--- a/sentry_streams/sentry_streams/adapters/arroyo/reduce_delegate.py
+++ b/sentry_streams/sentry_streams/adapters/arroyo/reduce_delegate.py
@@ -25,7 +25,6 @@ from sentry_streams.adapters.arroyo.rust_step import (
     RustOperatorDelegate,
     RustOperatorFactory,
 )
-from sentry_streams.config_types import ReduceConfig
 from sentry_streams.pipeline.message import Message, PyMessage
 from sentry_streams.pipeline.pipeline import Reduce
 
@@ -114,10 +113,9 @@ class ReduceDelegateFactory(RustOperatorFactory[TIn, TOut], Generic[TIn, TOut]):
     Creates a `ReduceDelegate`. This is the class to provide to the Rust runtime.
     """
 
-    def __init__(self, step: Reduce[Any, Any, Any], step_config: ReduceConfig) -> None:
+    def __init__(self, step: Reduce[Any, Any, Any]) -> None:
         super().__init__()
         self.__step = step
-        self.__step_config = step_config
 
     def build(self) -> ReduceDelegate[TIn, TOut]:
         retriever = OutputRetriever[TOut]()
@@ -125,7 +123,7 @@ class ReduceDelegateFactory(RustOperatorFactory[TIn, TOut], Generic[TIn, TOut]):
 
         return ReduceDelegate(
             build_arroyo_windowed_reduce(
-                self.__step.windowing(self.__step_config.batch_size),
+                self.__step.windowing,
                 self.__step.aggregate_fn,
                 retriever,
                 route,
--- a/sentry_streams/sentry_streams/adapters/arroyo/reduce_delegate.py
+++ b/sentry_streams/sentry_streams/adapters/arroyo/reduce_delegate.py
@@ -25,6 +25,7 @@ from sentry_streams.adapters.arroyo.rust_step import (
     RustOperatorDelegate,
     RustOperatorFactory,
 )
+from sentry_streams.config_types import ReduceConfig
 from sentry_streams.pipeline.message import Message, PyMessage
 from sentry_streams.pipeline.pipeline import Reduce
 
@@ -113,9 +114,10 @@ class ReduceDelegateFactory(RustOperatorFactory[TIn, TOut], Generic[TIn, TOut]):
     Creates a `ReduceDelegate`. This is the class to provide to the Rust runtime.
     """
 
-    def __init__(self, step: Reduce[Any, Any, Any]) -> None:
+    def __init__(self, step: Reduce[Any, Any, Any], step_config: ReduceConfig) -> None:
         super().__init__()
         self.__step = step
+        self.__step_config = step_config
 
     def build(self) -> ReduceDelegate[TIn, TOut]:
         retriever = OutputRetriever[TOut]()
@@ -123,7 +125,7 @@ class ReduceDelegateFactory(RustOperatorFactory[TIn, TOut], Generic[TIn, TOut]):
 
         return ReduceDelegate(
             build_arroyo_windowed_reduce(
-                self.__step.windowing,
+                self.__step.windowing(self.__step_config.batch_size),
                 self.__step.aggregate_fn,
                 retriever,
                 route,
--- a/sentry_streams/sentry_streams/pipeline/chain.py
+++ b/sentry_streams/sentry_streams/pipeline/chain.py
@@ -172,8 +172,15 @@ class Batch(
     Applier[Message[InputType], Message[MutableSequence[InputType]]],
     Generic[MeasurementUnit, InputType],
 ):
+    batch_size: MeasurementUnit
+
     def build_step(self, name: str, ctx: Pipeline, previous: Step) -> Step:
-        return BatchStep(name=name, ctx=ctx, inputs=[previous])
+        return BatchStep(
+            name=name,
+            ctx=ctx,
+            inputs=[previous],
+            batch_size=self.batch_size,
+        )
 
 
 @dataclass
--- a/sentry_streams/sentry_streams/pipeline/chain.py
+++ b/sentry_streams/sentry_streams/pipeline/chain.py
@@ -172,15 +172,8 @@ class Batch(
     Applier[Message[InputType], Message[MutableSequence[InputType]]],
     Generic[MeasurementUnit, InputType],
 ):
-    batch_size: MeasurementUnit
-
     def build_step(self, name: str, ctx: Pipeline, previous: Step) -> Step:
-        return BatchStep(
-            name=name,
-            ctx=ctx,
-            inputs=[previous],
-            batch_size=self.batch_size,
-        )
+        return BatchStep(name=name, ctx=ctx, inputs=[previous])
 
 
 @dataclass
--- a/sentry_streams/sentry_streams/pipeline/pipeline.py
+++ b/sentry_streams/sentry_streams/pipeline/pipeline.py
@@ -337,10 +337,13 @@ class Broadcast(WithInput):
             self.ctx.register_edge(self, branch_step)
 
 
+@dataclass
 class Reduce(WithInput, ABC, Generic[MeasurementUnit, InputType, OutputType]):
     """
     A generic Step for a Reduce (or Accumulator-based) operation
     """
+
+    app_config: Optional[Any] = field(default=None, init=False)
 
     @property
     @abstractmethod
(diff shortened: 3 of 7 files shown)


sentry_streams/sentry_streams/pipeline/chain.py
```diff
@@ -168,15 +168,8 @@ class Batch(
...
     Generic[MeasurementUnit, InputType],
 ):
-    batch_size: MeasurementUnit
-
     def build_step(self, name: str, ctx: Pipeline, previous: Step) -> Step:
-        return BatchStep(
-            name=name,
-            ctx=ctx,
-            inputs=[previous],
-            batch_size=self.batch_size,
```
fpacifici, line 178: The default can still be provided in code. Please do not remove this. 
WHat you provide in config is an override.

sentry_streams/sentry_streams/pipeline/pipeline.py
```diff
@@ -339,9 +339,8 @@ class Reduce(WithInput, ABC, Generic[MeasurementUnit, InputType, OutputType]):
     def group_by(self) -> Optional[GroupBy]:
         raise NotImplementedError()
 
-    @property
     @abstractmethod
-    def windowing(self) -> Window[MeasurementUnit]:
```
fpacifici, line 344: Why these changes?

sentry_streams/sentry_streams/adapters/arroyo/reduce_delegate.py
```diff
@@ -113,17 +114,18 @@ class ReduceDelegateFactory(RustOperatorFactory[TIn, TOut], Generic[TIn, TOut]):
...
+        self.__step_config = step_config
 
     def build(self) -> ReduceDelegate[TIn, TOut]:
         retriever = OutputRetriever[TOut]()
         route = Route(source="dummy", waypoints=[])
 
         return ReduceDelegate(
             build_arroyo_windowed_reduce(
-                self.__step.windowing,
+                self.__step.windowing(self.__step_config.batch_size),
```
fpacifici, line 128: This is the wrong abstraction. 
There are multiple types of window algorithms, not all of them have a batch size. 
It is only the `Batch` dataclass in pipelines for which batch size makes sense.

Ultimately I think the config should not go this deep in the stack. 
You should be able to transform the step object by overriding its configuration inside the adapter (rust_arroyo.py) so that the step object, at that point, already has the right value.

fpacifici (commented):
some high level comments

sentry_streams/sentry_streams/pipeline/chain.py
victoria-yining-huang: i think we should not allow `batch_size` to be passed in application code at all. I feel that many softwares we build (KCP, sentry-kube values files, consumer app cli default values/ deployment template default values), provide too many options to do the same thing. I can totally see that somebody looks at the config file and it looks right so they deploy, only finding out later that it's hardcoded in the application with a different value. 

But it's just my two cents this is not a hill I wanna die on

sentry_streams/sentry_streams/pipeline/pipeline.py
```diff
@@ -147,10 +147,20 @@ class Step:
...
         self.ctx.register(self)
 
+    def override_config(self, app_config: Any) -> Any:
+        """
+        A step's config in the application would be overriden by
+        its config value loaded from the file
+        """
+        value = self.config if self.config else app_config
+        self.config = value
+        return value
```
fpacifici, line 162: Why do you need to return it here ?

sentry_streams/sentry_streams/pipeline/pipeline.py
```diff
@@ -147,10 +147,20 @@ class Step:
 
     name: str
     ctx: Pipeline
+    config: Optional[Any] = field(default=None, init=False)
 
     def __post_init__(self) -> None:
         self.ctx.register(self)
 
+    def override_config(self, app_config: Any) -> Any:
```
fpacifici, line 155: Please provide a more precise type to `app_config`

sentry_streams/sentry_streams/pipeline/pipeline.py
```diff
@@ -147,10 +147,20 @@ class Step:
...
+        """
+        A step's config in the application would be overriden by
+        its config value loaded from the file
```
fpacifici, line 158: remove the mention to the file. This method does not need to know where the config overrides are coming from.

sentry_streams/sentry_streams/pipeline/pipeline.py
```diff
@@ -147,10 +147,20 @@ class Step:
 
     name: str
     ctx: Pipeline
+    config: Optional[Any] = field(default=None, init=False)
```
fpacifici, line 150: there is a simpler way to manage this optional config: delegate to the subclasses how to manage it  entirely.
Do not specify the existence of a `config` field here. Specifying it here  has one downside: you have to provide one also for steps that have no override ability.

Instead you can make the `override_config` method abstract or provide a `NotImplementedError` by default, so that only the steps that have a meaningful way to override parameters can implement it.

sentry_streams/sentry_streams/pipeline/pipeline.py
```diff
@@ -147,10 +147,20 @@ class Step:
...
 
     def __post_init__(self) -> None:
         self.ctx.register(self)
 
+    def override_config(self, app_config: Any) -> Any:
+        """
+        A step's config in the application would be overriden by
+        its config value loaded from the file
+        """
+        value = self.config if self.config else app_config
```
fpacifici, line 160: This approach would only allow you to override the entirety of the configuration. What would happen if a step has two parameters and you want to override only one?
Imagine if `Batch` had both a `batch_time` and a `batch_size`

sentry_streams/sentry_streams/pipeline/pipeline.py
```diff
@@ -329,11 +339,18 @@ def __post_init__(self) -> None:
...
 class Reduce(WithInput, ABC, Generic[MeasurementUnit, InputType, OutputType]):
     """
     A generic Step for a Reduce (or Accumulator-based) operation
+
+    app_config: this is the config being passed in from the application code
+        where the class is instantiated. This is the default value
+        when there is no overriding value defined in the config file.
     """
 
+    app_config: Optional[Any] = field(default=None, init=False)
```
fpacifici, line 352: Why does Reduce needs this attribute ?

fpacifici (changes requested):
Please see the comments in line. 
I think this approach is more complex than it needs to be. If you delegate to each step the override process you will not need an `app_config: Any` attribute that has unspecified behavior for all the subclasses that do not allow for override.

sentry_streams/sentry_streams/pipeline/pipeline.py
victoria-yining-huang: because in `runner.py` the `reduce` function is only aware the step is a reduce. `apply_config` is called there and it needs to pass in the `app_config` attribute. For a Batch step, `app_config` is the same as its `batch_size`, for other kinds of Reduce classes `app_config` will just be null

sentry_streams/sentry_streams/pipeline/pipeline.py
victoria-yining-huang: if you know a way to make an optional abstract method plz lmk it's a better way for sure

sentry_streams/sentry_streams/pipeline/pipeline.py
victoria-yining-huang: im not sure if the merge needs to be built right now

sentry_streams/sentry_streams/pipeline/pipeline.py
fpacifici: > So what do you mean "only the steps that have a meaningful way to override parameters can implement it."

You do not make it optional. You actually do not make it abstract. You provide a default implementation that does nothing. So you can always call override in all classes. Then you override that method in Batch and in all other classes that support override by adding the logic there.

victoria-yining-huang: nvm resolved on slack, does not return anything

sentry_streams/sentry_streams/pipeline/pipeline.py
```diff
@@ -151,6 +151,12 @@ class Step:
     def __post_init__(self) -> None:
         self.ctx.register(self)
 
+    def override_config(self, loaded_config: Mapping[str, Any] | None) -> Any:
```
fpacifici, line 154: I think this function should not return anything `-> None:` 

Also is there any difference for the function whether it is passed none or an empty Mapping? Supporting both hints that such a difference would exist. 
If they mean the same thing, please disallow one by only allow `Mapping[str, Any]`

sentry_streams/sentry_streams/pipeline/pipeline.py
```diff
@@ -391,7 +398,7 @@ class Batch(Reduce[MeasurementUnit, InputType, MutableSequence[InputType]]):
 
     # TODO: Use concept of custom triggers to close window
     # by either size or time
-    batch_size: MeasurementUnit
+    batch_size: Optional[MeasurementUnit] = None
```
fpacifici, line 401: This is not supposed to be optional. There must always be a default otherwise this introduces an implicit requirement that in some cases all regions must have an override for the batch_size.
If you require a default then we know there is a fallback value for any region that does not specify this field.

sentry_streams/sentry_streams/pipeline/pipeline.py
```diff
@@ -400,13 +407,24 @@ def group_by(self) -> Optional[GroupBy]:
 
     @property
     def windowing(self) -> Window[MeasurementUnit]:
+        assert (
+            self.batch_size is not None
+        ), f"{self.name} config must be set before windowing is accessed"
```
fpacifici, line 412: Not allowing None also allows you to remove this.

In general only support nullability when you need it. It will remove a lot of corner cases that you would have to test and maintain.

sentry_streams/tests/test_pipeline.py
```diff
@@ -368,3 +370,56 @@ def test_invalid_add() -> None:
 
     with pytest.raises(AssertionError):
         pipeline1.add(pipeline2)
+
+
+@pytest.mark.parametrize(
+    "loaded_batch_size, default_batch_size, expected",
+    [
+        pytest.param({"batch_size": 50}, 100, 50, id="Have both loaded and default values"),
+        pytest.param({"batch_size": 50}, None, 50, id="Only has loaded config file value"),
```
fpacifici, line 379: This scenario is not supposed to be allowed.

fpacifici (changes requested):
Please revert the nullability of the batch_size and see the other comments in line

sentry_streams/sentry_streams/pipeline/pipeline.py
victoria-yining-huang: done, `loaded_config` only allows `Mapping[str, Any]` now

victoria-yining-huang:
@fpacifici made default value mandatory

sentry_streams/tests/test_pipeline.py
victoria-yining-huang: removed

sentry_streams/sentry_streams/pipeline/pipeline.py
victoria-yining-huang: removed

fpacifici (approved):

PR #131, opened 2025-06-05
quick-fix: Ensure schema is retained after a batching step

ayirr7:
A really simple way to unblock the work where a batching step happens before parsing. I can also introduce the concept of `Schema` in this PR, where we have an enum of all the allowed schemas / topic names. 

We should not stick with this in the long term, and likely enforce all Accumulators/Reduce steps to output a `Tuple[MutableSequence[InputType], Schema]`. Making this is a more significant typing change as it involves making sure primitives like `FlatMap` are also able to handle that output type.

### DIFF
--- a/sentry_streams/sentry_streams/adapters/arroyo/reduce_delegate.py
+++ b/sentry_streams/sentry_streams/adapters/arroyo/reduce_delegate.py
@@ -11,6 +11,7 @@ from typing import (
     Tuple,
     TypeVar,
     Union,
+    cast,
 )
 
 from arroyo.processing.strategies.abstract import ProcessingStrategy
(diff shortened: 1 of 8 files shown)


sentry_streams/sentry_streams/adapters/arroyo/reduce_delegate.py
```diff
@@ -165,26 +166,43 @@ def submit(self, message: ArroyoMessage[Union[FilteredPayload, TIn]]) -> None:
...
+                    timestamp=timestamp,
+                    schema=schema,
+                )
+            else:
+                msg = PyMessage(
+                    payload=payload,
+                    headers=[],
+                    timestamp=timestamp,
+                    schema=None,
+                )
```
fpacifici, line 198: Please let's write unit tests for this logic (there is already a `test_reduce_delegate` you can add onto). 
As this is a bit of a hack, a good way to keep it under control is to have this behavior well tested so that when we refactor this we will not miss corner cases.

ayirr7: This test is already checking what we need here.  I made it a bit more "complex"

fpacifici (approved):

PR #132, opened 2025-06-05
Support parallelism in python

fpacifici:
Based on https://github.com/getsentry/streams/pull/120

Introduce a configurable Multi process step implementation for all the `Map` steps.
This relies on the Arroyo multi process step to run Pyhton logic in parallel. 
This logic is executed by the Rust runtime through the `ArroyoStrategyDelegate` that wraps an existing arroyo strategy.
The step can be configured through the pipeline config 
```
        transform:
          parallelism:
            multi_process:
              processes: 4
              batch_size: 1000
              batch_time: 0.2

```
The streaming application does not know anything about multi processing.

Limitations:
- It can only run one single step in each multi processing pool. We will support chaining multiple steps in the same process.


sentry_streams/sentry_streams/adapters/arroyo/multi_process_delegate.py
```diff
@@ -0,0 +1,179 @@
...
+
+    def build(
+        self,
+    ) -> ArroyoStrategyDelegate[
+        FilteredPayload | Message[TMapIn], FilteredPayload | Message[TMapOut]
+    ]:
+        retriever = OutputRetriever[Union[FilteredPayload, Message[TMapOut]]](mapped_msg_to_rust)
+
+        processor = RunTaskWithMultiprocessing(
+            partial(process_message, self.__step.resolved_function),
```
bmckerry, line 166: slightly confused as to how this is working - I know we're making a partial function of `process_message` where the `function` param is `Map.resolved_function()`, but `resolved_function()` doesn't take any args. Is this working because `resolved_function()` itself returns a `Callable`? How is `resolved_function(msg)` resolved?

bmckerry (approved):

sentry_streams/sentry_streams/adapters/arroyo/multi_process_delegate.py
fpacifici: That is correct `resolved_function`  returns a Callable, a Callable is an object itself (first class functions), thus it can be passed to another function which will call it.
We are passing it to process_message as the predefined argument of partial, which will then call it when it receives a message

PR #133, opened 2025-06-11
feat(rust_arroyo): add watermark step

bmckerry:
This PR adds a `Watermark` step to the rust arroyo consumer which is added at the top of the pipeline after the step that wraps messages in `RoutedValue`. It periodically emits a watermark message to downstream steps (currently just containing the timestamp of creation, will be updated in a future PR).

The goal of this PR is to just have a `Watermark` step that sends watermark messages downstream which are ignored by other steps (i.e. the watermarks don't actually do anything). This is to prevent the PR from being completely unreadable. 

Watermark PRs:
- [ ] pr1 (this PR): introduce `Watermark` step and `WatermarkMessage` type
- [ ] pr2: watermark message body - tracking offsets for messages sent before watermark messages, etc
- [ ] pr3: fix reduce/broadcast/router - each needs to have special behaviour when they see a watermark msg
- [ ] pr4: commit policy - building a custom commit policy that only sends a `CommitRequest` when a watermark msg is seen by the policy


sentry_streams/src/watermark.rs
```diff
@@ -0,0 +1,144 @@
...
+    pub route: Route,
+    pub period: u64,
+    last_timestamp: u64,
+}
+
+impl Watermark {
+    /// A strategy that periodically sends watermark messages.
+    /// This strategy is added as the first step in a consumer.
+    /// The Arroyo adapter commit step only commits once it recives a watermark
+    /// message with a specific committable from all branches in a consumer.
```
fpacifici, line 29: This explains the struct, so it should be put on top of the struct declaration not in the impl.

sentry_streams/src/watermark.rs
```diff
@@ -0,0 +1,144 @@
...
+    pub fn new(
+        next_step: Box<dyn ProcessingStrategy<RoutedValue>>,
+        route: Route,
+        period: Option<u64>,
```
fpacifici, line 33: Why does this need to be optional ?

sentry_streams/src/watermark.rs
```diff
@@ -0,0 +1,144 @@
...
+
+    fn send_watermark_msg(&mut self) -> Result<(), SubmitError<RoutedValue>> {
+        let timestamp = current_epoch();
+        let watermark_msg = RoutedValue {
+            route: self.route.clone(),
+            payload: RoutedValuePayload::WatermarkMessage(WatermarkMessage::new(timestamp as f64)),
+        };
+        self.last_timestamp = timestamp;
+        self.next_step
+            .submit(Message::new_any_message(watermark_msg, BTreeMap::new()))
```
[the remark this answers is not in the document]
fpacifici: @john-z-yang  is right. The last timestamp should be updated only if the `submit` is successful. If not we need to attempt submitting again the same message (with the same timestamp - does not need to be the same instance) at the next call of poll.

sentry_streams/src/watermark.rs
```diff
@@ -0,0 +1,144 @@
...
+impl ProcessingStrategy<RoutedValue> for Watermark {
+    fn poll(&mut self) -> Result<Option<CommitRequest>, StrategyError> {
+        self.next_step.poll()
+    }
+
+    fn submit(&mut self, message: Message<RoutedValue>) -> Result<(), SubmitError<RoutedValue>> {
+        if self.should_send_watermark_msg() {
+            self.send_watermark_msg()?;
+        }
+        self.next_step.submit(message)
```
fpacifici, line 73: This algorithm will not work. 
If there is no more message passed through we do not send the next watermark. 
The watermark should be sent on poll rather than submit as poll is called periodically no matter if there are new messages from kafka.

sentry_streams/src/watermark.rs
```diff
@@ -0,0 +1,144 @@
...
+
+    fn join(&mut self, timeout: Option<Duration>) -> Result<Option<CommitRequest>, StrategyError> {
+        self.next_step.join(timeout)?;
+        if self.should_send_watermark_msg() {
+            match self.send_watermark_msg() {
+                //TODO: not sure how to handle MessageRejected here
+                Ok(..) => (),
+                Err(..) => (),
+            };
+        }
```
fpacifici, line 88: During join you have to send a watermark no matter what. This is basically a guarantee to flush. 
Moreover you will have to induce all the steps downstream to flush their state. This can be done in one of two ways:
- rely on calling join on the downstream steps. The stateful downstream steps would have to preserve all the watermarks to know what they have to flush everything and forward the watermarks when they receive a call to join.
- make the watermark carry the "flush" information, which means an attribute on the Watermark message that signals this watermark was sent after a join call thus no more messages are expected.

fpacifici (changes requested):
Please see the comments inline

sentry_streams/src/messages.rs
```diff
@@ -78,6 +78,21 @@ pub fn headers_to_sequence(
     Ok(seq.unbind())
 }
 
+/// WatermarkMessages are sent by the Watermark step periodically to ensure
+/// the consumer's commit policy functions even when the consumer strategy
+/// contains multiple branching routes.
+#[derive(Debug, Copy, Clone)]
+#[pyclass]
```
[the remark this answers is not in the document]
untitaker: i can answer that, you need to be able to do `isinstance(message, WatermarkMessage)` from python code to forward the watermark msg from rust to python and back, in-order. the timestamp can be f64 because python doesn't need to access it. see also TODO down below, the python side wasn't completed yet

sentry_streams/src/messages.rs
```diff
@@ -286,6 +328,7 @@ impl Into<PyStreamingMessage> for Py<PyAny> {
 pub enum StreamingMessage {
     PyAnyMessage { content: PyAnyMessage },
     RawMessage { content: RawMessage },
+    WatermarkMessage { content: WatermarkMessage },
```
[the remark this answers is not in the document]
bmckerry: I'm still not sure when a `StreamingMessage` would be used so I added `WatermarkMessage (I know it's for situations where we don't need to call python code, just not sure what use we would have for that). I'll remove it from here

sentry_streams/src/messages.rs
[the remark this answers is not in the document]
bmckerry: Like @untitaker said, I made it a pyclass so that the message can be passed to python code. Updating python code so it actually does something with a watermark message will be done in a future PR.

sentry_streams/src/watermark.rs
bmckerry: Discussed in slack, we'll be going with option 1 `rely on calling join on the downstream steps`. The code for this will be added in the future PR that updates the reduce step to handle watermark msgs.

sentry_streams/src/filter_step.rs
```diff
@@ -173,6 +182,16 @@ mod tests {
...
+
+            let watermark_val = RoutedValue {
+                route: Route::new(String::from("source"), vec![]),
+                payload: RoutedValuePayload::WatermarkMessage(WatermarkMessage::new(0.)),
+            };
+            let watermark_msg = Message::new_any_message(watermark_val, BTreeMap::new());
+            let watermark_res = strategy.submit(watermark_msg);
+            assert!(watermark_res.is_ok());
+            let watermark_messages = submitted_watermarks_clone.lock().unwrap();
+            assert_eq!(watermark_messages.len(), 1);
```
bmckerry, line 194: In all tests regarding `WatermarkMessages` I just assert the length of the recorded watermark list as the payload of the messages is a timestamp which is inconsistent between tests.

In the next PR when I update the payload of `WatermarkMessage` to be a committable I'll update all related tests to assert on the actual `WatermarkMessage` payload.

sentry_streams/src/consumer.rs
```diff
@@ -345,7 +357,8 @@ mod tests {
 
             let submitted_messages = Arc::new(Mutex::new(Vec::new()));
             let submitted_messages_clone = submitted_messages.clone();
-            let next_step = FakeStrategy::new(submitted_messages, false);
+            let submitted_watermarks = Arc::new(Mutex::new(Vec::new()));
+            let next_step = FakeStrategy::new(submitted_messages, submitted_watermarks, false);
```
fpacifici, line 361: Please consider passing a watermark mesage in this test and validate it ends up in `submitted_watermarks`

sentry_streams/src/filter_step.rs
```diff
@@ -40,29 +40,35 @@ impl ProcessingStrategy<RoutedValue> for Filter {
...
+                                let filtered = boolean.is_truthy(py).unwrap();
+                                Ok(filtered)
+                            }
+                            Err(e) => Err(e),
+                        }
+                    })
+                }
+                RoutedValuePayload::WatermarkMessage(..) => {
+                    unreachable!("Watermark message trying to be passed to filter function.")
                 }
```
fpacifici, line 69: If I remember correctly we could not have a proper `match` block here because we could not forward the message from the `RoutedValuePayload::PyStreamingMessage`. 

For "proper match block" I mean
```
match message.payload() {
    WatermarkMessage => {}
    PyStreamingMessage => match {
          PyStreamingMessage::PyAnyMessage =>
          PyStreamingMessage::RawMessage =>
```

If that's the case, please add a comment explaining why we cannot do that.

sentry_streams/src/filter_step.rs
fpacifici: That's alright

sentry_streams/src/python_operator.rs
```diff
@@ -451,7 +457,8 @@ class RustOperatorDelegateFactory:
 
             let submitted_messages = Arc::new(RawMutex::new(Vec::new()));
             let submitted_messages_clone = submitted_messages.clone();
-            let next_step = FakeStrategy::new(submitted_messages, false);
+            let submitted_watermarks = Arc::new(RawMutex::new(Vec::new()));
+            let next_step = FakeStrategy::new(submitted_messages, submitted_watermarks, false);
```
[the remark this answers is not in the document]
fpacifici: I think this was not addressed.

sentry_streams/src/watermark.rs
```diff
@@ -0,0 +1,121 @@
...
+        let mut watermark = Watermark::new(
+            Box::new(next_step),
+            Route {
+                source: String::from("source"),
+                waypoints: vec![],
+            },
+            10,
+        );
+        watermark.last_timestamp = 0;
+        let _ = watermark.poll();
```
fpacifici, line 118: test that after line 119, if you do not reset last_timestamp, if you submit a message and poll you would not see another watermark in the next step.

fpacifici (approved):
Please address the missing comments before merging

sentry_streams/src/filter_step.rs
bmckerry: I made this be `unreachable` because if the message was actually a `RoutedValuePayload::WatermarkMessage`, it would be caught by this if block at the top of the function:
```
if self.route != message.payload().route || message.payload().payload.is_watermark_msg() {
            self.next_step.submit(message)
        }
```
I think this is the behaviour we want from the filter step (forwarding watermark messages without submitting them to python), should I still add a comment?

fpacifici: > I made this be unreachable because if the message was actually a RoutedValuePayload::WatermarkMessage, it would be caught by this if block at the top of the submit method:

That would be a runtime check.
You have to have two branches in your code that can semantically expect a watermark message. One is invalid and has to crash. A bug that makes that code reachable would crash the consumer.

Doing this via pattern matching would make the check a compile time. It would not be possible to end in a panic situation. Thus it is desirable to push to compile time all checks we can reasonably push to compile time. They make it less likely to crash the consumer. 

Thus my question

> If I remember correctly we could not have a proper match block here because we could not forward the message from the RoutedValuePayload::PyStreamingMessage.

The question is whether making this a compile time check is viable.

PR #140, opened 2025-06-16
feat(chain): add ParquetSerializer

victoria-yining-huang:
Serializing to Parquet functionality should be provided by the streaming platform.
This PR also adds some generic streaming data types so we can abstract away third party libraries from users.

the `parquet_serializer.py` example worked locally


sentry_streams/pyproject.toml
```diff
@@ -15,6 +15,11 @@ dependencies = [
     "pyyaml>=6.0.2",
     "jsonschema>=4.23.0",
     "sentry-kafka-schemas>=1.2.0",
+    "polars==1.30.0",
+    "pyarrow==19.0.0",
+    "pyarrow-stubs==19.0",
+    "pandas==2.2.3",
```
fpacifici, line 21: Please do not pin versions. 
polars>=1.30.0 
pyarrow>=19.0.0
pandas>=2.2.3

If you pin versions and the client of the library requires a different version they would not be able to use this library.

fpacifici: Also you should not need `pyarrow-stubs` in the main dependencies.
Please add them in the `dev` section

fpacifici, line 21: Do you need pandas only for tests?
If yes please do not add this dependency. We should minimize the dependencies we import in a library. That makes it easier to use in other code bases.

Option 1: do without pandas 
Option 2: add it only to the dev section we use for tests.

fpacifici (changes requested):
Please see my comment on the dependencies

sentry_streams/sentry_streams/pipeline/msg_codecs.py
```diff
@@ -45,3 +58,92 @@ def custom_serializer(obj: Any, dt_format: Optional[str] = None) -> str:
...
+
+def _map_arrow_to_polars_schema(schema: Sequence[Tuple[str, PADataType]]) -> PolarsSchema:
+    return pl.Schema(
+        {field_schema[0]: _map_arrow_to_polars_dtype(field_schema[1]) for field_schema in schema}
+    )
+
+
+def _get_parquet(msg: Message[Any], schema_fields: PolarsSchema) -> bytes:
+    df = pl.DataFrame(
+        [i for i in msg.payload if i is not None],
```
fpacifici, line 134: How does this work? `Message[Any]` is not known to be iterable to the type checker.

sentry_streams/sentry_streams/pipeline/chain.py
```diff
@@ -167,6 +173,20 @@ def build_step(self, name: str, ctx: Pipeline, previous: Step) -> Step:
         )
 
 
+@dataclass
+class ParquetSerializer(Applier[Message[TIn], bytes], Generic[TIn]):
```
fpacifici, line 177: I am not sure the type of the message is right. Isn't the input message supposed to be a `Message[pyarrow.Table]` or `Message[pyarrow.Array]`.

Or is it a Sequence of python dictionary as it seems from the unit test ?

fpacifici (changes requested):

sentry_streams/pyproject.toml
victoria-yining-huang: removed dependency versions, moved pyarrow-stubs to dev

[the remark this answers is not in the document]
fpacifici: > i still need these packages in the streams repo for the tests? 
Test run on the dev feature  https://github.com/getsentry/streams/blob/main/sentry_streams/pyproject.toml#L26-L35. It has more dependencies.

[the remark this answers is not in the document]
victoria-yining-huang: ok got it

sentry_streams/sentry_streams/pipeline/msg_codecs.py
```diff
@@ -54,3 +62,92 @@ def custom_serializer(obj: Any, dt_format: Optional[str] = None) -> str:
...
+        schema=schema_fields,
+    )
+    buffer = io.BytesIO()
+    df.write_parquet(buffer, compression="lz4", statistics=False, use_pyarrow=False)
```
victoria-yining-huang, line 142: the old implementation uses snappy

sentry_streams/sentry_streams/pipeline/chain.py
victoria-yining-huang: no, the input message is not `Message[pyarrow.Table]` or `Message[pyarrow.Array]`. It is `Message[Sequence[Mapping[str, Any]]]` or `Message[Mapping[str, Any]]`

sentry_streams/pyproject.toml
```diff
@@ -33,6 +34,7 @@ dev = [
     "maturin>=1.5.1",
     "types-PyYAML>=6.0.12",
     "types-jsonschema>=4.23.0",
+    "polars",
```
fpacifici, line 37: No need to add it here. You already added it to the main dependencies.

sentry_streams/sentry_streams/pipeline/chain.py
```diff
@@ -190,6 +193,23 @@ def build_step(self, name: str, ctx: Pipeline, previous: Step) -> Step:
         )
 
 
+@dataclass
+class ParquetSerializer(Applier[Message[TIn], bytes]):
+    schema_fields: Mapping[str, Any]
```
fpacifici, line 198: I guess this is a `Mapping[str, StreamsDataType]` rather than `Mapping[str, Any]`

sentry_streams/sentry_streams/pipeline/msg_codecs.py
```diff
@@ -54,3 +72,51 @@ def custom_serializer(obj: Any, dt_format: Optional[str] = None) -> str:
...
+    assert isinstance(polars_schema, PolarsSchema)
+    return polars_schema
+
+
+def parquet_serializer(
+    msg: Message[Any],
+    schema_fields: Mapping[str, DataType],
+    compression: Optional[ParquetCompression] = "snappy",
+) -> bytes:
+    assert _validate_schema(schema_fields)
```
fpacifici, line 120: This method is called on every message. Please do here only what you need to do for every message. Validating that the schema is structured correctly does not need to be done here. Please do it when you instantiate the `ParquetSerializer`

sentry_streams/sentry_streams/pipeline/msg_codecs.py
```diff
@@ -54,3 +72,51 @@ def custom_serializer(obj: Any, dt_format: Optional[str] = None) -> str:
...
+    polars_schema = _resolve_polars_schema(schema_fields)
```
fpacifici, line 121: Same, please resolve the schema to polars when you instantiate the `ParquetSerailizer` and reuse it.

sentry_streams/sentry_streams/pipeline/msg_codecs.py
```diff
@@ -54,3 +72,51 @@ def custom_serializer(obj: Any, dt_format: Optional[str] = None) -> str:
 
     serializer = partial(custom_serializer, dt_format=dt_format)
     return json.dumps(payload, default=serializer).encode("utf-8")
+
+
+ParquetCompression = Literal["lz4", "uncompressed", "snappy", "gzip", "lzo", "brotli", "zstd"]
+
+
+def _get_parquet(
+    msg: Message[Any],
```
fpacifici, line 81: What's the type of the message?  This is not supposed to typecheck as you use the payload of the message as an iterator below. If you expect a sequence then please provide the right type here.

sentry_streams/sentry_streams/pipeline/chain.py
```diff
@@ -190,6 +193,23 @@ def build_step(self, name: str, ctx: Pipeline, previous: Step) -> Step:
         )
 
 
+@dataclass
+class ParquetSerializer(Applier[Message[TIn], bytes]):
```
fpacifici, line 197: IS this the right type of message ? (`TIn` ) 
It seems to me that `parquet_serializer` expects a sequence

sentry_streams/sentry_streams/pipeline/msg_codecs.py
```diff
@@ -54,3 +72,51 @@ def custom_serializer(obj: Any, dt_format: Optional[str] = None) -> str:
...
+            raise TypeError(f"Field {k} has type {dtype} and it is not a valid Streams DataType.")
+        return True
+
+    return all(_check_dtype(k, dtype) for k, dtype in schema.items())
+
+
+def _resolve_polars_schema(schema_fields: Mapping[str, Any]) -> PolarsSchema:
+    resolved_schema = {key: dtype.resolve() for key, dtype in schema_fields.items()}
+    polars_schema = pl.Schema(resolved_schema)
+    assert isinstance(polars_schema, PolarsSchema)
```
fpacifici, line 111: what else can this be ?

sentry_streams/sentry_streams/pipeline/msg_codecs.py
```diff
@@ -54,3 +72,51 @@ def custom_serializer(obj: Any, dt_format: Optional[str] = None) -> str:
...
+
+
+def _get_parquet(
+    msg: Message[Any],
+    schema_fields: PolarsSchema,
+    compression: Union[ParquetCompression, None] = "snappy",
+) -> bytes:
+    df = pl.DataFrame([i for i in msg.payload if i is not None], schema=schema_fields)
+    buffer = io.BytesIO()
+    assert compression is not None
```
fpacifici, line 87: Why do you accept a `Union[ParquetCompression, None]` if you assert that `compression` cannot be none ? Just do not take None and you do not need the assertion.

sentry_streams/sentry_streams/pipeline/msg_codecs.py
```diff
@@ -54,3 +72,51 @@ def custom_serializer(obj: Any, dt_format: Optional[str] = None) -> str:
...
+    def _check_dtype(k: str, dtype: StreamsDataType) -> bool:
+        if isinstance(dtype, Struct):
+            return all(_check_dtype(k, field.dtype) for field in dtype.fields)
+        elif isinstance(dtype, List):
+            return _check_dtype(k, dtype.inner)
+        elif isinstance(dtype, Field):
+            return _check_dtype(k, dtype.dtype)
+
+        if not isinstance(dtype, StreamsDataType):
+            raise TypeError(f"Field {k} has type {dtype} and it is not a valid Streams DataType.")
```
fpacifici, line 102: I am not sure what the goal of this function is. You are just checking that the types of all the values of `schema` are StreamingDataType. IT seems to me they already cannot be anything else by construction thus a validation is not needed. 
Or am I missing something? Could you please provide an example of how an invalid schema could be built ?

sentry_streams/tests/pipeline/test_msg_codecs.py
```diff
@@ -78,3 +88,136 @@ def test_msg_no_found_schema() -> None:
...
+
+    result = parquet_serializer(msg, schema)
+
+    assert isinstance(result, bytes)
+
+    df = pl.read_parquet(BytesIO(result))
+    assert df.shape == (2, 8)
+
+    expected_df = pl.DataFrame(
+        [
```
fpacifici, line 157: If this is identical to the one above, please put the list in a constant and use it both to build the parquet dataframe and to verify that, otherwise we will eventually update one forgetting the other.

sentry_streams/sentry_streams/pipeline/msg_codecs.py
victoria-yining-huang: without this validation, if the schema uses a non-streams DataType (such as python data types) the error would be 
```
E       AttributeError: type object 'str' has no attribute 'resolve'

sentry_streams/pipeline/datatypes.py:157: AttributeError
```
with validation:

```
E           TypeError: Field org_id has type <class 'int'> and it is not a valid Streams DataType.

sentry_streams/pipeline/msg_codecs.py:102: TypeError
```

it gives a more informative error imo

sentry_streams/sentry_streams/pipeline/msg_codecs.py
victoria-yining-huang: removed

sentry_streams/sentry_streams/pipeline/msg_codecs.py
victoria-yining-huang: never none now

sentry_streams/sentry_streams/pipeline/msg_codecs.py
victoria-yining-huang: that only works if every message is confirmed to have the same schema. How is that guaranteed?

victoria-yining-huang: would somehow messages from two different sources end up in the same batch?

sentry_streams/sentry_streams/pipeline/chain.py
victoria-yining-huang: done

sentry_streams/sentry_streams/pipeline/msg_codecs.py
fpacifici: Why would somebody use a non-streams DataType ?

sentry_streams/sentry_streams/pipeline/msg_codecs.py
fpacifici: I don't think you need a generic here at all.
A Generic is supposed to allow you to build a class/function/algorithm that works independently from the type provided. 
Example:
Imagine a List: it accumulates elements of a type, it allows you to add and get elements of that type. The type does not matter regarding the behavior of the List. So you could use Any as type but then the type checker would not be able to enforce that the type you get from the `get` method is the same as the type you use in the `add` method. That's where generic is for. It enforces that `add` and `get` deal with the same type. 

In this case you know that the message can only be of one of two specific types. The code of this method is not agnostic to the specific type. 
Line https://github.com/getsentry/streams/pull/140/files#diff-348980743026bbba64bf8ea5a9e6558721d5d4f05f46d3284260512d01437f00R83
is dependent on the type:
- payload has to be an iterable `[i for i in msg.payload` 
- the content of each entry in msg.payload has to be something you can pass to `pl.DataFrame()`  Which is not `Any` https://docs.pola.rs/api/python/stable/reference/dataframe/index.html#dataframe. 

So the type of the payload of your message cannot be `Any` nor it has to be Generic. From the behavior of the function it should be something like `Message[Iterable[FrameInitTypes`]]`. I am not sure that is what you expected though.

sentry_streams/sentry_streams/pipeline/msg_codecs.py
victoria-yining-huang: resolve in the ParquetSerializer constructor

sentry_streams/sentry_streams/pipeline/msg_codecs.py
victoria-yining-huang: this is removed now. Validating schema will happen at type checking not at run time

sentry_streams/pyproject.toml
victoria-yining-huang: removed

sentry_streams/sentry_streams/pipeline/msg_codecs.py
victoria-yining-huang: resolved on a call. Removed generic

john-z-yang (approved):

PR #144, opened 2025-06-17
feat: Introduce object file generator for GCSSink

ayirr7:
Allows users to define and pass in a function reference to a function for generating object file names for the GCS sink. This allows for dynamically-generated object files, for examples, file names with the current datetime during the GCS write.


sentry_streams/src/gcs_writer.rs
```diff
@@ -18,8 +18,9 @@ use sentry_arroyo::processing::strategies::run_task_in_threads::TaskRunner;
 use sentry_arroyo::types::Message;
 pub struct GCSWriter {
     client: Client,
-    url: String,
+    bucket: String,
     route: Route,
+    object_generator: Py<PyAny>,
```
fpacifici, line 23: I think you can use `Py<PyFunction>`

sentry_streams/src/gcs_writer.rs
```diff
@@ -59,16 +56,39 @@ impl GCSWriter {
...
+            object_generator,
+        }
     }
 }
 
+fn object_gen_fn(object_generator: Py<PyAny>, py: Python<'_>) -> PyResult<String> {
+    let res: Py<PyAny> = object_generator.call0(py)?;
+    let output: String = res.extract(py)?;
+
+    Ok(output)
```
fpacifici, line 74: Doesn't this work as  

```
 let res: Py<PyAny> = object_generator.call0(py)?;
 res.extract(py)
```

PR #146, opened 2025-06-18
fix(examples): Migrate billing example

evanh:
This changes the billing example to use the new chain syntax instead of the old method. It also
changes the signature of the OutcomesBuffer.

There is also a test that puts a message through the pipeline and verifies the output.


evanh:
The test I added doesn't actually work, it fails with this stack trace:

```
Traceback (most recent call last):
  File "/Users/evanhicks/code/getsentry/streams/sentry_streams/.venv/lib/python3.13/site-packages/arroyo/processing/processor.py", line 334, in run
    self._run_once()
    ~~~~~~~~~~~~~~^^
  File "/Users/evanhicks/code/getsentry/streams/sentry_streams/.venv/lib/python3.13/site-packages/arroyo/processing/processor.py", line 405, in _run_once
    self.__message = self.__consumer.poll(timeout=1.0)
                     ~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^
  File "/Users/evanhicks/code/getsentry/streams/sentry_streams/.venv/lib/python3.13/site-packages/arroyo/backends/kafka/consumer.py", line 427, in poll
    raise OffsetOutOfRange(str(error))
arroyo.errors.OffsetOutOfRange: KafkaError{code=_AUTO_OFFSET_RESET,val=-140,str="fetch failed due to requested offset not available on the broker: Broker: Offset out of range (broker 1001)"}
```

Not sure why this error is happening in the runtime code.

sentry_streams/sentry_streams/examples/billing.py
```diff
@@ -1,50 +1,87 @@
...
-    inputs=[map],
-    window=reduce_window,
-    aggregate_func=OutcomesBuffer,
-    aggregate_backend=KVAggregationBackend(),  # NOTE: Provided by the platform
+pipeline = (
+    streaming_source(
+        name="myinput",
+        stream_name="events",
+    )
+    .apply("mymap", Map(function=build_outcome))
```
fpacifici, line 81: Please use a `Parser` applier rather than a custom map. https://getsentry.github.io/streams/build_pipeline.html#sentry_streams.pipeline.chain.Parser

We have the outcome schema in the schema registry so it should just work https://github.com/getsentry/sentry-kafka-schemas/blob/main/schemas/outcomes.v1.schema.json

As people will refer to these example, we should follow the practices we want to recommend

sentry_streams/tests/test_pipeline_w_messages.py
```diff
@@ -0,0 +1,91 @@
...
+    "retention_days": 90,
+    "type": "c",
+    "value": 1,
+}
+
+
+def run_pipeline(pipeline: Pipeline) -> None:
+    with open("./sentry_streams/deployment_config/simple_map_filter.yaml", "r") as f:
+        environment_config = yaml.safe_load(f)
+    print(":HERE", environment_config)
```
fpacifici, line 43: was this forgotten ?

fpacifici:
> arroyo.errors.OffsetOutOfRange: KafkaError{code=_AUTO_OFFSET_RESET,val=-140,str="fetch failed due to requested offset not available on the broker: Broker: Offset out of range (broker 1001)"}

This means that either:
- the consumer was started with auto-offset-reset = error and there was no offset in kafka (new topic or consumer group? )
- the consumer was started with auto-offset-reset = error and the committed offset was already collected and dropped 24 hours after it was created.

I'd verify the auto offset reset parameter you are passing to kafka. Initial offset reset parameter can be provided to the consumer in the yaml config file.https://github.com/getsentry/streams/blob/main/sentry_streams/sentry_streams/adapters/arroyo/rust_arroyo.py#L89-L94

evanh:
I'll remove the test from this, I was trying to do some kind of integration test. I'll put that in a separate PR.

fpacifici (approved):

PR #147, opened 2025-06-18
fix(examples): Rewrite examples using the chain style API

evanh:
Rewrite the examples to use the chain style API. The billing example is migrated in a separate PR. Remove some of the helper files since they weren't being used outside of the example and weren't very complicated.

Also skip the `alerts.py` file since it uses a FlatMap, and that has to be implemented in a separate PR.


evanh:
I couldn't get the typing to work for the `FlatMap`, I get these errors:

```
sentry_streams/sentry_streams/examples/alerts.py:28: error: Argument 2 to "apply" of "ExtensibleChain" has incompatible type "FlatMap[Never, Never]"; expected "Applier[Message[Event], Message[Never]]"  [arg-type]
sentry_streams/sentry_streams/examples/alerts.py:28: error: Argument "function" to "FlatMap" has incompatible type "Callable[[Message[Event]], list[TimeSeriesDataPoint]]"; expected "Callable[[Message[MutableSequence[Never]]], Never] | str"  [arg-type]
sentry_streams/sentry_streams/examples/alerts.py:34: error: Argument 2 to "apply" of "ExtensibleChain" has incompatible type "Reducer[int, list[TimeSeriesDataPoint], p95AlertData | CountAlertData]"; expected "Applier[Message[Never], Message[p95AlertData | CountAlertData]]"  [arg-type]
Found 3 errors in 1 file (checked 7 source files)
```

sentry_streams/sentry_streams/examples/alerts.py
```diff
@@ -1,69 +1,42 @@
...
-    name="kafkasink",
-    ctx=pipeline,
-    inputs=[map_str],
-    stream_name="transformed-events",
+pipeline = (
+    streaming_source(
+        name="myinput",
+        stream_name="events",
+    )
+    .apply("mymap", Map(function=build_event))
```
fpacifici, line 22: Let's turn this into a parser. The logic will be slightly different because, currently, this pipeline uses a custom event class rather than the one defined in the schema registry, but it should be a small change.

sentry_streams/sentry_streams/examples/broadcast.py
```diff
@@ -1,50 +1,40 @@
...
+# Create the source stream
+source = streaming_source("myinput", stream_name="events")
 
-goodbye_map = Map(
-    name="goodbye_map",
-    ctx=pipeline,
-    inputs=[map],
-    function=BroadcastFunctions.goodbye_map,
+# Apply the no-op map transformation
+mapped: ExtensibleChain[Message[str]] = source.apply(
```
fpacifici, line 10: Was this typehint needed ?

sentry_streams/sentry_streams/examples/broadcast_fn.py
```diff
@@ -11,15 +13,15 @@ class BroadcastFunctions:
     """
 
     @staticmethod
-    def no_op_map(value: str) -> str:
-        return value
+    def no_op_map(value: Message[bytes]) -> str:
```
fpacifici, line 16: thanks

sentry_streams/sentry_streams/examples/events.py
```diff
@@ -78,16 +81,18 @@ def __init__(self) -> None:
         self.alert_type: str
         self.alert_id: int
 
-    def add(self, value: TimeSeriesDataPoint) -> Self:
-        if value.alert_type == "count":
-            self.count += 1
-            self.alert_type = value.alert_type
-            self.alert_id = value.alert_id
+    def add(self, message: Message[list[TimeSeriesDataPoint]]) -> Self:
```
fpacifici, line 84: Not sure I get why we end up having a `list[TimeSeriesDataPoint]`.
IF I remember correctly, we map, then flatmap to produce messages each of which contains a `TimeSeriesDataPoint`. Is the list needed because flatMap does not work ?

sentry_streams/sentry_streams/examples/spans_buffer.py
```diff
@@ -1,29 +1,93 @@
...
+
+
+def build_span(value: Message[bytes]) -> Span:
+    """
+    Build a Span object from a JSON str
+    """
+
+    d: dict[str, Any] = json.loads(value.payload)
+
+    return Span(d["span_id"], d["trace_id"], d["duration"], d["timestamp"])
```
fpacifici, line 37: Let's use the schema from sentry-kafka-schemas and a Parser rather than a custom event type.

fpacifici:
```
sentry_streams/sentry_streams/examples/alerts.py:28: error: Argument 2 to "apply" of "ExtensibleChain" has incompatible type "FlatMap[Never, Never]"; expected "Applier[Message[Event], Message[Never]]"  [arg-type]
sentry_streams/sentry_streams/examples/alerts.py:28: error: Argument "function" to "FlatMap" has incompatible type "Callable[[Message[Event]], list[TimeSeriesDataPoint]]"; expected "Callable[[Message[MutableSequence[Never]]], Never] | str"  [arg-type]
sentry_streams/sentry_streams/examples/alerts.py:34: error: Argument 2 to "apply" of "ExtensibleChain" has incompatible type "Reducer[int, list[TimeSeriesDataPoint], p95AlertData | CountAlertData]"; expected "Applier[Message[Never], Message[p95AlertData | CountAlertData]]"  [arg-type]
Found 3 errors in 1 file (checked 7 source files)
```

Yeah it seems the flatMap implementation is just plain broken and we are noticing it only now because the old `Step` api basically allows any type while the Chain api does some more reasonable type checking. 

The issue is the following:
- The signature of the applier for FlatMap is partially coherent 
- - The coherent part: https://github.com/getsentry/streams/blob/main/sentry_streams/sentry_streams/pipeline/chain.py#L97. The input type is a sequence of `Message` the output is a `Message`. This may seem counterintuitive but i is correct as the output type is meant to represent the input type of the following step. It does not tell anything about the cardinality of the output. FlatMap takes a batch and produces the messages contained in the batch one by one. The "one by one" is correctly not captured in the output type.
- - The incoherent part: The input for a flatMap does not need to be a batch. The signature there seems to imply it has to be a batch. While we could produce many alerts from a single event https://github.com/getsentry/streams/blob/main/sentry_streams/sentry_streams/examples/events.py#L120. So the input type should be less restrictive.
- This is the part that is plain wrong https://github.com/getsentry/streams/blob/main/sentry_streams/sentry_streams/pipeline/chain.py#L98-L100. The flatMap function is unbatching a batch. It should return a `Iterator[TOut]` not just one `TOut` (or any other representation of the sequence we like). 
- The `materialize_alerts` function in the example is conceptually correct but has the wrong type. https://github.com/getsentry/streams/blob/main/sentry_streams/sentry_streams/examples/events.py#L120 It takes an Event (see the "incoherent part" above) and returns a Generator. Which means we can iterate on the produced messages one by one, which is correct.

So I think there are two fixes to make. And I would do them in a separate PR:
1. Change the return type of the function provided to the FlatMap to be iterable as we should get multiple messages out. Iterable seems a better idea than a Sequence.
2. Make the input type more generic and do not impose a Batch structure.

What do you think ?

sentry_streams/sentry_streams/examples/broadcast.py
evanh: Yep.

sentry_streams/sentry_streams/examples/spans_buffer.py
evanh: I thought we were trying to move away from sentry-kafka-schemas?

evanh:
@fpacifici It mostly makes sense, just to clarify:

> FlatMap takes a batch and produces the messages contained in the batch one by one. The "one by one" is correctly not captured in the output type.

This is confusing me a bit, I thought FlatMap was allowing a 1->N style of processing. So one message in can produce one or more messages out. If this is the case I agree that it shouldn't enforce a Batch style input.

I'm reluctant to have the output be a Generator however. I would think we'd want a more concrete data structure, in case the next step is written in Rust.

sentry_streams/sentry_streams/examples/alerts.py
evanh: What's wrong with using a map?

evanh:
I removed the changes to FlatMap from this PR, I'll migrate that example in a separate PR.

sentry_streams/sentry_streams/examples/spans_buffer.py
fpacifici: Yes, but the replacement will have an interface that will be very close to that one. It should be a quick swap, we can use it till we replace it.

fpacifici:
> @fpacifici Could you take another look at this? The flatmap changes went into a separate PR, and I'd like to get this merged soon so it doesn't conflict too much with https://github.com/getsentry/streams/pull/152

See my comment on `broadcast.py`. I don't think that pipeline works.

sentry_streams/sentry_streams/examples/broadcast.py
```diff
@@ -1,50 +1,40 @@
...
+    "goodbye_map",
+    Map(
+        function=BroadcastFunctions.goodbye_map,
+    ),
+).sink(
+    "goodbye_sink",
+    StreamSink(
+        stream_name="transformed-events-2",
+    ),
 )
```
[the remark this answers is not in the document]
evanh: This API is completely broken then. The `broadcast()` function takes a sequence of chains, but how do I create those chains? I can't create a chain without a `Source`, and my two branches here don't have sources.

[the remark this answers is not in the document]
evanh: Looking further I missed the `segment` function. I updated the code to use that.

fpacifici (approved):

PR #149, opened 2025-06-19
ref(logging): make traced_with_gil display callsite filename and lineno

john-z-yang:


sentry_streams/src/transformer.rs
```diff
@@ -106,7 +106,7 @@ mod tests {
 
     fn import_py_dep(module: &str, attr: &str) {
         let stmt = format!("from {} import {}", module, attr);
-        traced_with_gil("import_py_dep", |py| {
+        traced_with_gil!(|py: Python<'_>| {
```
john-z-yang, line 109: I'm not sure why rustc is not smart enough to infer the types here 😕

sentry_streams/src/utils.rs
```diff
@@ -1,28 +1,64 @@
...
+///
+/// let py_err = ...;
+/// traced_with_gil!(Duration::from_secs(10), |py| {
+///     py_err.print(py)
+/// }).unwrap();
+/// ```
+///
+macro_rules! traced_with_gil {
+    ($function: expr) => {
+        traced_with_gil!(std::time::Duration::from_secs(1), $function)
```
ayirr7, line 23: (beginner question) Is this basically invoking the macro? So it will resolve to the second branch?

sentry_streams/src/utils.rs
```diff
@@ -1,28 +1,64 @@
...
+use std::{
+    thread,
+    time::{Duration, Instant},
+};
 use tracing::warn;
 
-// Use this wrapper instead of directly using with_gil()
-pub fn traced_with_gil<F, R>(label: &str, function: F) -> R
+#[doc(hidden)]
+pub(crate) fn __traced_with_gil<F, R>(label: &str, warn_threshold: Duration, function: F) -> R
```
ayirr7, line 45: What does the `pub(crate)` here do exactly?

ayirr7 (approved):

sentry_streams/src/utils.rs
john-z-yang: It just makes it available for the things inside this create. It looks like this codebase will be also a library, I don't think it should be made public to the people using the library.

As for why it is public in the first place (because ideally this is only used by the macro), there doesn't seem to be a way to create a private helper for the macro since it's just syntactic abstraction and it disappears when it's expanded.

PR #151, opened 2025-06-19
fix(flatmap): Change the FlatMap signature to match usage

evanh:
The FlatMap takes a single message as input (i.e. a batch) and outputs 1->N messages. Currently have
the output type be a `Iterable[TOut]` and the next step would be required to iterate the output from
the FlatMap.

Note that this is a temporary measure: the API currently doesn't have the capability of handling
1->N style primitives.

### DIFF
--- a/sentry_streams/sentry_streams/examples/alerts.py
+++ b/sentry_streams/sentry_streams/examples/alerts.py
@@ -1,13 +1,15 @@
+from sentry_kafka_schemas.schema_types.events_v1 import InsertEvent
+
 from sentry_streams.examples.events import (
     AlertsBuffer,
     GroupByAlertID,
     build_alert_json,
-    build_event,
     materialize_alerts,
 )
 from sentry_streams.pipeline import (
     FlatMap,
     Map,
+    Parser,
     Reducer,
     streaming_source,
 )
@@ -19,7 +21,7 @@ pipeline = (
         name="myinput",
         stream_name="events",
     )
-    .apply("mymap", Map(function=build_event))
+    .apply("parser", Parser(msg_type=InsertEvent))
     # We add a FlatMap so that we can take a stream of events (as above)
     # And then materialize (potentially multiple) time series data points per
     # event. A time series point is materialized per alert rule that the event
--- a/sentry_streams/sentry_streams/examples/events.py
+++ b/sentry_streams/sentry_streams/examples/events.py
@@ -1,18 +1,12 @@
 import json
 from dataclasses import dataclass
-from typing import Any, Self, Union
+from datetime import datetime
+from typing import Generator, Self, Union
+
+from sentry_kafka_schemas.schema_types.events_v1 import InsertEvent
 
 from sentry_streams.pipeline.function_template import Accumulator, GroupBy
 from sentry_streams.pipeline.message import Message
-
-
-@dataclass
-class Event:
-    project_id: int
-    latency: int
-    timestamp: int
-    tags: list[str]
-    value: int
 
 
 @dataclass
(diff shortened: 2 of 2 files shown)


sentry_streams/sentry_streams/examples/alerts.py
```diff
@@ -1,69 +1,42 @@
...
-    name="kafkasink",
-    ctx=pipeline,
-    inputs=[map_str],
-    stream_name="transformed-events",
+pipeline = (
+    streaming_source(
+        name="myinput",
+        stream_name="events",
+    )
+    .apply("mymap", Map(function=build_event))
```
fpacifici, line 22: Please consider replacing this with a Parser so we are consistent in our example in what the pattern to parse events should be.

sentry_streams/sentry_streams/pipeline/chain.py
```diff
@@ -94,10 +95,13 @@ def build_step(self, name: str, ctx: Pipeline, previous: Step) -> Step:
 
 
 @dataclass
-class FlatMap(Applier[Message[MutableSequence[TIn]], Message[TOut]], Generic[TIn, TOut]):
-    function: Union[
-        Callable[[Message[MutableSequence[TIn]]], TOut], str
-    ]  # TODO: Consider making this type an Iterable rather than MutableSequence
+class FlatMap(Applier[Message[TIn], Message[Iterable[TOut]]], Generic[TIn, TOut]):
```
fpacifici, line 98: I don't think this should type check correctly 

```
class FlatMap(Applier[Message[TIn], Message[Iterable[TOut]]], Generic[TIn, TOut])
```

While the function that does the transformation in a flatMap returns an `Iterable[TOut]`, the next step in the chain will take in a `Message[TOut]` not a `Message[Iterable[TOut]]`. 

The second generic parameter of the `Applier` function does not represent the type returned by the transformation function. It represents, instead, the type of each message produced by the step, which has to coincide with the input type of the following step in the chain.

This is what allows us to ensure via type checking that you can only chain a step that is compatible with the previous one. Mypy can check that the input type of a step is equal to the output type of the previous step. 

As an example:
let's say we have a source -> parser -> flatMap -> serializer -> sink
- Source would be a source[Message[bytes]]
- parser would be be a Map[Message[bytes], Messag[MyEventbatch]]
- flatMap would be a FlatMap[Messag[MyEventBatch], Message[MyEvent]]
- serializer would be a Map[Message[MyEvent], message[bytes]]

On
```
FlatMap[Messag[MyEventBatch], Message[MyEvent]]
```

It is true that the flatmap function provided by the user is 
```
Callable[[Messag[MyEventBatch]], Iterable[MyEvent]]
```
but the following step takes one `Message[MyEvent]` at a time, so its input type is `Message[MyEvent]`. Which means the output type of FlatMap has to be `Message[MyEvent]`.
Which is expected as the FlatMap step has to take the iterable, unpack it and call `submit` multiple times on the following step.

evanh: > FlatMap step has to take the iterable

This keeps coming up: why does FlatMap have to take an iterable? Why couldn't any single message result in many messages?

fpacifici: Sorry I was not precise on this 

> Which is expected as the FlatMap step has to take the iterable, unpack it and call submit multiple times on the following step.

What I actually meant is that FlatMap has to take the message the flatMap function can unpack, not necessarily a Python Iterable. 
FlatMap does not need to take a python Iterable.

evanh: Got it. Fixed the typing.

fpacifici (approved):

PR #153, opened 2025-06-20
ref: Handle overrides in GCSSink from config

ayirr7:
Addresses https://github.com/getsentry/streaming-planning/issues/182. GCS Sink takes 2 configuration parameters
- bucket name (is static)
- object file names (can be dynamically generated)

This PR adds support for handling any overrides that may be provided in the config file for a specific pipeline. The rule of thumb is that any params defined in the config file will override any params as defined in the pipeline.

### DIFF
--- a/sentry_streams/sentry_streams/adapters/arroyo/rust_arroyo.py
+++ b/sentry_streams/sentry_streams/adapters/arroyo/rust_arroyo.py
@@ -167,14 +167,10 @@ class RustArroyoAdapter(StreamAdapter[Route, Route]):
                 bucket = (
                     step.bucket if not sink_config.get("bucket") else str(sink_config.get("bucket"))
                 )
-                object_generator = (
-                    step.object_generator
-                    if not sink_config.get("object_file")
-                    else lambda: str(sink_config.get("object_file"))
-                )
             else:
                 bucket = step.bucket
-                object_generator = step.object_generator
+
+            object_generator = step.object_generator
 
             self.__consumers[stream.source].add_step(
                 RuntimeOperator.GCSSink(route, bucket, object_generator)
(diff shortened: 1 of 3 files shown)


sentry_streams/sentry_streams/adapters/arroyo/rust_arroyo.py
```diff
@@ -163,9 +163,23 @@ def sink(self, step: Sink, stream: Route) -> Route:
...
         if isinstance(step, GCSSink):
+            if sink_config := self.steps_config.get(step.name):
+                bucket = (
+                    step.bucket if not sink_config.get("bucket") else str(sink_config.get("bucket"))
+                )
+                object_generator = (
+                    step.object_generator
+                    if not sink_config.get("object_file")
+                    else lambda: str(sink_config.get("object_file"))
+                )
```
fpacifici, line 174: I think we should fail if there is a mismatch between object name generator and statically defined object name.
If the application code defines a generator, we should not allow the config to override it with a statically defined object name. That would likely break the semantics of the application.

fpacifici (approved):
Please address the comment before merging.

sentry_streams/sentry_streams/adapters/arroyo/rust_arroyo.py
ayirr7: spunds good

fpacifici: no, sorry I misunderstood that. 
IF we always provide a callable, then there should be no way to override that via config.

ayirr7: Ok, I'll remove the ability to override the generators, but keep the ability to override the bucket name

PR #154, opened 2025-06-20
ref(rust): fixup clippy lint

john-z-yang:
A few minor refactor
- Mark unused code with `#[allow(unused)]`
- Remove usage of `&` in places where the value is immediately auto-dereferenced
- Mark commented tests as #[ignore]
- Adds cargo clippy to rust CI

### DIFF
--- a/sentry_streams/src/messages.rs
+++ b/sentry_streams/src/messages.rs
@@ -129,6 +129,7 @@ pub struct PyAnyMessage {
     pub schema: Option<String>,
 }
 
+#[allow(dead_code)]
 pub fn into_pyany(py: Python<'_>, message: PyAnyMessage) -> PyResult<Py<PyAnyMessage>> {
     // Converts a PyAnyMessage instantiated outside of Python memory into a
     // Py<PyAnyMessage> that can be used in Python code.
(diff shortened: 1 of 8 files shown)


sentry_streams/src/operators.rs
```diff
@@ -112,7 +112,7 @@ pub fn build(
                 route.clone(),
                 next,
                 concurrency_config,
-                &bucket,
+                bucket,
```
ayirr7, line 115: I'm curious how you were able to remove the `&` here?

john-z-yang: Because `bucket` is already a `&String`, `&String` coerces to &str

sentry_streams/src/messages.rs
```diff
@@ -129,6 +129,7 @@ pub struct PyAnyMessage {
     pub schema: Option<String>,
 }
 
+#[allow(dead_code)]
```
fpacifici, line 132: https://github.com/getsentry/streams/blob/4ee04065320076478b07d265620286c0d3ed700b/sentry_streams/src/test_operators.rs#L39

I think this is not dead code

sentry_streams/src/transformer.rs
```diff
@@ -135,90 +135,93 @@ mod tests {
...
-    //             transform.submit(message).unwrap_err()
-    //         else {
-    //             panic!("Expected SubmitError::InvalidMessage")
-    //         };
-
-    //         assert_eq!(partition, Partition::new(Topic::new("topic"), 2));
-    //         assert_eq!(offset, 10);
-    //     });
-    // }
+    #[ignore]
```
fpacifici, line 138: Is anybody fixing these tests ?

fpacifici (approved):

PR #157, opened 2025-06-24
ref(rust): Parse and handle errors when interfacing with python code in route and filter

john-z-yang:
- Unify rust → python facilities in `transform`, `route` and `filter` with `try_apply_py`. When calling into python with `try_apply_py` the return value will be `Result<T, ApplyError>`, where `ApplyError` can be either `InvalidMessage`, which is meant to signal to the event loop to not retry the message, or `ApplyFailed`, which indicates some ordinary errors. `try_apply_py` also handles logging python stack traces in event of an error.
- Add conversion for `ApplyError::InvalidMessage` → Arroyo runtime's `MessageRejected`
- Implements `Into` traits for `PyStreamingMessage` → `Py<PyAny>`, since we're doing the `match` + `clone_ref` over and over again.

### DIFF
--- a/sentry_streams/src/callers.rs
+++ b/sentry_streams/src/callers.rs
@@ -1,37 +1,29 @@
-use crate::messages::PyStreamingMessage;
-use crate::utils::traced_with_gil;
-use pyo3::prelude::*;
+use pyo3::{import_exception, prelude::*, types::PyTuple};
 
-/// Executes a Python callable with an Arroyo message containing Any and
-/// returns the result.
-pub fn call_python_function(
-    callable: &Py<PyAny>,
-    message: &PyStreamingMessage,
-) -> Result<PyStreamingMessage, PyErr> {
-    Ok(traced_with_gil!(|py| match message {
-        PyStreamingMessage::PyAnyMessage { ref content } => {
-            callable.call1(py, (content.clone_ref(py),))
-        }
+import_exception!(sentry_streams.pipeline.exception, InvalidMessageError);
 
-        PyStreamingMessage::RawMessage { ref content } => {
-            callable.call1(py, (content.clone_ref(py),))
-        }
-    })?
-    .into())
+pub type ApplyResult<T> = Result<T, ApplyError>;
+
+#[derive(Debug, PartialEq)]
+pub enum ApplyError {
+    InvalidMessage,
+    ApplyFailed,
 }
 
-/// Executes a Python callable with an Arroyo message containing Any and
-/// returns the result.
-pub fn call_any_python_function(
+pub fn try_apply_py<'py, N>(
+    py: Python<'py>,
     callable: &Py<PyAny>,
-    message: &PyStreamingMessage,
-) -> Result<Py<PyAny>, PyErr> {
-    traced_with_gil!(|py| match message {
-        PyStreamingMessage::PyAnyMessage { ref content } => {
-            callable.call1(py, (content.clone_ref(py),))
-        }
-        PyStreamingMessage::RawMessage { ref content } => {
-            callable.call1(py, (content.clone_ref(py),))
+    args: N,
+) -> ApplyResult<Py<PyAny>>
+where
+    N: IntoPyObject<'py, Target = PyTuple>,
+{
+    callable.call1(py, args).map_err(|py_err| {
+        py_err.print(py);
+        if py_err.is_instance(py, &py.get_type::<InvalidMessageError>()) {
+            ApplyError::InvalidMessage
+        } else {
+            ApplyError::ApplyFailed
         }
     })
 }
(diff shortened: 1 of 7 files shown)


sentry_streams/src/messages.rs
```diff
@@ -311,6 +311,17 @@ impl RoutedValuePayload {
...
 }
 
+impl From<&PyStreamingMessage> for Py<PyAny> {
+    fn from(value: &PyStreamingMessage) -> Self {
+        traced_with_gil!(|py| {
+            match &value {
+                PyStreamingMessage::PyAnyMessage { content } => content.clone_ref(py).into_any(),
+                PyStreamingMessage::RawMessage { content } => content.clone_ref(py).into_any(),
+            }
+        })
```
john-z-yang, line 321: I am wondering if we should just use `.clone().into_any()` here and avoid acquiring GIL, then creates a separate impl that takes the GIL token and use `clone_ref` in places where we already have the GIL.

sentry_streams/src/filter_step.rs
```diff
@@ -42,53 +43,30 @@ impl ProcessingStrategy<RoutedValue> for Filter {
...
-                    unreachable!("Watermark message trying to be passed to filter function.")
-                }
-            };
+            return self.next_step.submit(message);
+        }
+
+        let RoutedValuePayload::PyStreamingMessage(ref py_streaming_msg) =
+            message.payload().payload
+        else {
+            unreachable!("Watermark message trying to be passed to filter function.")
```
ayirr7, line 52: From a quick look, I wasn't able to tell why this is unreachable?

sentry_streams/src/callers.rs
```diff
@@ -1,52 +1,76 @@
...
-            callable.call1(py, (content.clone_ref(py),))
-        }
-        PyStreamingMessage::RawMessage { ref content } => {
-            callable.call1(py, (content.clone_ref(py),))
+pub fn try_apply_py<'py, N>(py: Python<'py>, callable: &Py<PyAny>, args: N) -> ApplyResut<Py<PyAny>>
+where
+    N: IntoPyObject<'py, Target = PyTuple>,
+{
+    callable.call1(py, args).map_err(|py_err| {
+        py_err.print(py);
```
evanh, line 18: Why do you do this?

sentry_streams/src/messages.rs
evanh: Can you safely clone the object without acquiring the GIL?

sentry_streams/src/callers.rs
john-z-yang: We want to print the stack trace of the error so we see it in the logs. I think another way to do this is to extract the error into a rust string, and integrate it with the tracing crate, but that has a decent amount of runtime cost.

sentry_streams/src/filter_step.rs
```diff
@@ -42,53 +43,30 @@ impl ProcessingStrategy<RoutedValue> for Filter {
...
-                            partition: inner.partition,
-                            offset: inner.offset,
-                        }))
-                    }
-                    InnerMessage::AnyMessage(inner) => {
-                        panic!("Unexpected message type: {:?}", inner)
-                    }
-                },
-            }
+        match (res, &message.inner_message) {
```
ayirr7, line 64: This is more ergonomic, thanks

sentry_streams/src/filter_step.rs
john-z-yang: `RoutedValuePayload` can be either be a `PyStreamingMessage` or `WatermarkMessage`, the else case is for when `message.payload().payload` is not a `PyStreamingMessage`

bmckerry: To be clear, it's unreachable because if the message is a watermark it just gets submitted immediately here: https://github.com/getsentry/streams/pull/157/files/d10264d7020c70812b646fc49e030a27f2128578?diff=split&w=0#diff-87f1346e603e2cbc5c6ac02b302d3061883ff9dec3010621eaacd215c79d3204R45-R46

sentry_streams/src/messages.rs
john-z-yang: No, looking at [their source code](https://pyo3.rs/main/doc/src/pyo3/instance.rs#1746-1754). I think if we call `clone` on a py<T>, it will try to acquire the GIL inside the function and increment the ref count. Which is basically what we're doing right now.

bmckerry (approved):
thank you for this, I'll try to remember how to use `From` because that's really cool

sentry_streams/src/messages.rs
john-z-yang: There's a small runtime cost associated `traced_with_gil`. So, if someone calls `into` while they have the gil, it will do some duplicate work because in the body this calls `traced_with_gil` again.

I think what I meant with my earlier comment is that, I was considering another function that takes a `py` in addition to PyStreamingMessage which runs `clone_ref` without calling `traced_with_gil` again

bmckerry: yeah, from what I understand the only ways to clone a python object via pyo3 are:
- get the GIL, copy the object by reference (increment the reference counter) via `clone_ref()`
- get the GIL, call out to the python `deep_copy()` function to actually clone the object value

PR #158, opened 2025-06-25
fix(uv): Pin python versions

untitaker:
fix a few things:

* `.python-version` was ineffective because it is not located next to `pyproject.toml` (uv wants it that way)
* when running `cargo test`, it would not have the right pythonpath -- port over the workaround from snuba
* setup-python does not have to specify a version, we can make it read the same python-version as uv does


ayirr7:
Would you be able to add a description on what issues this change fixes and how?

untitaker:
I was advised by @joshuarli to not try and use the same exact python builds in CI and dev, because it's too difficult to use indygreg python builds in CI.

so i think this is the next-best thing.

sentry_streams/src/testutils.rs
```diff
@@ -0,0 +1,14 @@
+use pyo3::prelude::*;
+
+pub fn initialize_python() {
```
bmckerry, line 3: could we move this function into `test_operators.rs` (or move everything from there into here)?
It already has a few test-related utility functions in it

bmckerry (approved):

sentry_streams/src/testutils.rs
untitaker: i just renamed test_operators to testutils... now it makes sense again (to me)

evanh (approved):

PR #164, opened 2025-06-27
feat(API): add Unfold generic step for 1:n message relationships

bmckerry:


sentry_streams/src/broadcaster.rs
```diff
@@ -0,0 +1,265 @@
...
+            self.retry_pending_message(message, identifier)
+        } else {
+            self.submit_to_next_step(message, identifier)
+        }
+    }
+}
+
+impl ProcessingStrategy<RoutedValue> for Broadcaster {
+    fn poll(&mut self) -> Result<Option<CommitRequest>, StrategyError> {
+        self.flush_pending()?;
```
bmckerry, line 137: This is a situation I'm not sure about:
- a watermark message goes thru the broadcast step with 2 downstream branches, the first copy receives a `MessageRejected` backpressure error and goes into the pending buffer
- before the watermark message is retried, `poll()` is called which flushes the pending buffer, sending the previously failed message copy downstream 
  - **commit step receives watermark from branch 1**
- the failed message is retried and successfully passes through the broadcast step, sending 2 more copies of the message downstream 
  - **commit step receives watermark from branches 1 and 2**

In the above situation we end up with 3 watermarks reaching the commit step when the commit step only expects 2 (since there's 2 downstream routes), meaning if we're just counting watermarks we'll commit after we receive the 2nd watermark copy from branch 1 without receiving the watermark copy from branch 2.

Questions:
- can arroyo call `poll()` between when a message receives `MessageRejected` and when that message is retried?
- does this mean that in the custom commit step, we'll need to ensure we receive a watermark with a specific `committable` payload routed for each possible branch, and not just count the number of watermarks with a given `committable` payload?

evanh (approved):
I think it looks good!

sentry_streams/src/broadcaster.rs
```diff
@@ -0,0 +1,262 @@
...
+    }
+
+    fn submit(&mut self, message: Message<RoutedValue>) -> Result<(), SubmitError<RoutedValue>> {
+        if self.route != message.payload().route {
+            return self.next_step.submit(message);
+        }
+        let unfolded_messages = generate_broadcast_messages(&self.downstream_branches, message);
+        for msg in unfolded_messages {
+            self.handle_submit(msg)?;
+        }
```
fpacifici, line 145: I don't think this works as expected.
As this piece of code does not flush the pending messages, if there are pending messages, you would submit the newer message before the pending messages (that are older) thus sending messages out of order, which is wrong.

Please ensure this case is covered by your unit tests.

sentry_streams/src/broadcaster.rs
fpacifici: I am not sure I understand the problem. 
Why would we ever send a second copy of the watermark ? 
If we try to send a watermark, route 1 works and route 2 receives a MessageRejected:
- The next poll should only re-send route 2 watermark.
- route 1 watermak should not be sent again as you do for any other message.

Are we not using the message identifier to avoid sending another copy of the same watermark ?

sentry_streams/src/broadcaster.rs
```diff
@@ -0,0 +1,262 @@
...
+            let _ = step.submit(message.clone());
+            assert_eq!(step.pending_messages.len(), 1);
+            let actual_messages = submitted_messages_clone.lock().unwrap();
+            assert_messages_match(
+                py,
+                vec![
+                    "test_message".into_py_any(py).unwrap(),
+                    "test_message".into_py_any(py).unwrap(),
+                ],
+                actual_messages.deref(),
```
fpacifici, line 249: This is not enough to test the functionality. It does not check messages are sent on all routes.

PR #165, opened 2025-06-27
Fix the process to append steps to the pipeline

fpacifici:
https://github.com/getsentry/streams/pull/135 broke the way we append
Map steps to the main pipeline. In the last version I think I only
tested it end to end with parallel maps. So single threaded maps were
nbot been added to the chain, that included parsers and serializers.

This caused runtime error where messages were not parsed nor serialized
thus the issues in trying to send PyAnyMessage to Kafka directly.

This fixes the broken condition so that all steps are added and adds
some logs to ensure we can catch these in an easier way


victoria-yining-huang (approved):
i understand the code and the fix. Please run this on simple map filter before merging

fpacifici:
> Please run this on simple map filter before merging

PR #168, opened 2025-06-27
fix(ci): Also add release tags

untitaker:
I can't easily find the most recent version from github releases, fix
that

### DIFF
--- a/sentry_streams/.craft.yml
+++ b/sentry_streams/.craft.yml
@@ -7,6 +7,7 @@ changelogPolicy: auto
 preReleaseCommand: bash ../scripts/bump-version.sh
 releaseBranchPrefix: release-sentry-streams
 targets:
+  - name: github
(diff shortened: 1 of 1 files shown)


sentry_streams/.craft.yml
```diff
@@ -7,6 +7,7 @@ changelogPolicy: auto
 preReleaseCommand: bash ../scripts/bump-version.sh
 releaseBranchPrefix: release-sentry-streams
 targets:
+  - name: github
```
fpacifici, line 10: This was intentional, as we have two libraries in the same repo we cannot use the github releasing as that only allow one current release per repo.

Now, we never started cutting releases for sentry_flink, and we are not likely to start soon, we are likely to move it to another repo  if we get there. So in the end no big deal.

fpacifici (approved):

PR #173, opened 2025-07-04
feat(api): Remove the chains.py API

evanh:
Now that the Pipeline class can directly achieve the same API as the chains.py file, use that in all cases and remove the chains.py file as tech debt.

Also add a `route`, `broadcast` and `segment` function to the pipeline to emulate the API exactly.

Follow up to https://github.com/getsentry/streams/pull/162


sentry_streams/tests/adapters/arroyo/test_consumer.py
```diff
@@ -228,7 +235,7 @@ def test_multiple_routes(
     consumer.add_step(
         MapStep(
             route=Route(source="source1", waypoints=[]),
-            pipeline_step=cast(Map, router_pipeline.steps["decoder"]),
+            pipeline_step=cast(Map, cast(ComplexStep, router_pipeline.steps["decoder"]).convert()),
```
untitaker, line 238: why is this double-cast necessary?

evanh: Because `MapStep` expects specifically a `Map` step. The output of `convert` is a general `Step`.

fpacifici (changes requested):
Please address the comment on the routing function return type before moving on

PR #180, opened 2025-07-15
ref(pipeline): Restore type checking between steps

untitaker:
Restore type checking between steps and add a new test that ensures broken pipelines throw mypy errors.

Adding generics in the pipeline step forces me to add them everywhere, but we don't really need the type safety further down in adapters. There, we use Any because there's also no other way.


evanh (approved):
Thanks for fixing this Markus. I don't love the `[Any]` everywhere but I don't see a better way.

PR #183, opened 2025-07-24
Decouple the translator from the runner

fpacifici:
This is a sketch of how we could decouple the runner from the DSL
and from the code that processes the pipeline before handing it
to the runtime.

Decoupling is meant to allow us to have a fully rust runner without
having to port DSL and transaltion to Rust. Having a fully Rust
runner can help us to have a simpler way to build hybrid applicaiton
and to use the same runner to run fully Rust applications.

Running a streaming application consists of three components:
1. The DSL used to represent the pipeline structure. We have a Pyhton
   one today. Likely we will have more, though they will all produce
   the same graph structure.
2. Translation. This is a process where the pipeline is handed step
   by step to an Adapter to run it. In this process we are also
   transforming some steps from higher level primitives into a
   small number of low level ones. Example here is where a `Parser`
   step becomes a `Map` step. 
3. Runtime. This is the component that runs the consumer built via
   the adapter at step 2.

The idea in this PR ios that we can run 1 and 2 in Python, then produce
a low level description of the pipeline that we dump in a config file,
then we hand the config file to the Rust runner which can autonomously 
run the pipeline.

So:
1. Introduce the an abstraction layer in between `rust_arroyo.py` and the runtime.
   This can be found as `ArroyoConsumer` in dumping_consumer.py.
2. There are two implementations: one does what the system was doing before
   and creates an Arroyo Consumer. The other accumulates all the parameters
   and then dumps them in a YAML file. 
3. Add a new runner that loads the YAML file that mimics what a rust runner would do.
4. The runner still has to instantiate the python pipeline in hybrid applications
   to get a reference to the python functions to run. Though this is not 
   needed for rust only applications.

Caveats and different directions:
1. nothing works, this is just a sketch.
2. I think we may get away with something simpler. In hybrid applications 
   we are likely to start the rust runner from a python interpreter anyway.
   this allows the application developer to package the application logic
   in the same crate as the runner. Though, since we are starting it from
   a pyhton interpreter, we do not need to dump the application first.
   we can still run the translator.
   The dump would still be needed for fully rust applications.

### DIFF
--- a/sentry_streams/sentry_streams/adapters/arroyo/dumping_consumer.py
+++ b/sentry_streams/sentry_streams/adapters/arroyo/dumping_consumer.py
@@ -1,0 +1,203 @@
+from abc import ABC, abstractmethod
+from typing import Any, Callable, Mapping, Sequence
+
+from sentry_streams.adapters.arroyo.routes import Route
+from sentry_streams.config_types import MultiProcessConfig
+from sentry_streams.pipeline.message import Message
+from sentry_streams.pipeline.pipeline import Broadcast, Filter, GCSSink, Reduce, Router
+from sentry_streams.rust_streams import PyKafkaProducerConfig
+
+
+class ArroyoConsumer(ABC):
+    def __init__(self) -> None:
+        self.__chain: Sequence[Mapping[str, Any]] = []
+
+    @abstractmethod
+    def add_gcs_sink(self, step_name: str, route: Route, bucket: str, step: GCSSink) -> None:
+        raise NotImplementedError
+
+    @abstractmethod
+    def add_stream_sink(
+        self, step_name: str, route: Route, stream_name: str, kafka_config: PyKafkaProducerConfig
+    ) -> None:
+        raise NotImplementedError
+
+    @abstractmethod
+    def add_map(self, step_name: str, route: Route) -> None:
+        raise NotImplementedError
+
+    @abstractmethod
+    def add_filter(self, step_name: str, route: Route, step: Filter[Any]) -> None:
+        raise NotImplementedError
+
+    @abstractmethod
+    def add_multiprocess_map(
+        self,
+        step_name: str,
+        route: Route,
+        config: MultiProcessConfig,
+        func: Callable[[Message[Any]], Message[Any]],
+    ) -> None:
+        raise NotImplementedError
+
+    @abstractmethod
+    def add_broadcast(self, step_name: str, route: Route, step: Broadcast[Any]) -> None:
+        raise NotImplementedError
+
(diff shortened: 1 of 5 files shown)


sentry_streams/sentry_streams/adapters/arroyo/dumping_consumer.py
fpacifici, line 1: This file introduces the abstraction between the rust arroyo adapter and the lower level description of the pipeline.
One implementation creates a rust consumer, the other just dumps all the details in a sequence, which can be wrote in yaml and passed to the rust runner for pure rust applications.

sentry_streams/sentry_streams/runner_config.py
fpacifici, line 1: This runner mimics what the rust one would do:
1. load the yaml containing the pipeline descriptor produced by DumpingConsumer.
2. still loads the python DSL pipeline if the pipeline is hybrid. That is because the reference to the python application logic to run is there. For hybrid pipelines we can assume we start from a python CLI, though the difference from the previous attempt is that now the streams crate is packaged together with the rust application logic.
3. It would start the rust runner passing the yaml and the references to the needed python functions. (see TODO)

sentry_streams/sentry_streams/adapters/arroyo/dumping_consumer.py
```diff
@@ -0,0 +1,203 @@
...
+            "route": {"source": route.source, "waypoints": route.waypoints},
+        }
+        self.steps.append(step_description)
+
+    def add_multiprocess_map(
+        self,
+        step_name: str,
+        route: Route,
+        config: MultiProcessConfig,
+        func: Callable[[Message[Any]], Message[Any]],
```
john-z-yang, line 130: TODO: maybe register this function into a global registry so that the rust code can retrieve it when parsing the result of the dumping consumer

sentry_streams/sentry_streams/adapters/arroyo/rust_consumer.py
```diff
@@ -0,0 +1,116 @@
...
+            RuntimeOperator.PythonAdapter(
+                RustRoute(route.source, route.waypoints), ReduceDelegateFactory(step)
+            )
+        )
+
+    def run(self) -> None:
+        self.__consumer.run()
+
+    def dump(self) -> Any:
+        raise NotImplementedError
```
john-z-yang, line 116: TODO: have this thing return the list of runtime operators, and have the rust binary compose it. This way, should we want rust functions, since the runner is in rust the rust functions can be statically linked. We can represent the rust counterparts in the Operators enum.

PR #184, opened 2025-07-25
ref(metrics): Add metrics for tracking pipeline status

evanh:
This attempts to wrap steps with automatic metrics, so that users can see information about the
input/output of their steps as well as duration timings. The metrics are only implemented for the
rust arroyo runtime, as there is no easy way to generically add metrics for all runtimes.

This also attempts to initialize the metrics for arroyo, so that arroyo will log its metrics if that
run time is being used.


sentry_streams/sentry_streams/adapters/arroyo/rust_arroyo.py
```diff
@@ -64,6 +70,27 @@
...
+    if message_size:
+        metrics.increment("streams.pipeline.input.bytes", tags=tags, value=message_size)
+    return time.time()
+
+
+def output_metrics(name: str, error: bool, start_time: float, message_size: int | None) -> None:
+    metrics = get_metrics()
+    tags = {"step": name}
+    if error:
+        metrics.increment("streams.pipeline.errors", tags=tags)
```
bmckerry, line 86: Would it be worth recording the type of error that occurred as a tag?

sentry_streams/sentry_streams/metrics/metrics.py
```diff
@@ -0,0 +1,118 @@
+from __future__ import annotations
+
+from abc import abstractmethod
+from typing import Mapping, Optional, Protocol, Union, runtime_checkable
+
+from datadog import DogStatsd  # type: ignore[import-not-found]
```
bmckerry, line 6: why do we need this ignore?

sentry_streams/src/metrics.rs
```diff
@@ -0,0 +1,58 @@
...
+// Initialize metrics if config is provided
+pub fn configure_metrics(metric_config: Option<PyMetricConfig>) {
+    if let Some(ref metric_config) = metric_config {
+        println!(
+            "Initializing metrics with host: {}:{}",
+            metric_config.datadog_host(),
+            metric_config.datadog_port()
+        );
+
+        match UdpMetricSink::new(metric_config.datadog_host(), metric_config.datadog_port()) {
```
bmckerry, line 36: This can just be a TODO, but I feel like we'll want to decouple `configure_metrics` from datadog a bit since this will break on any other metrics backend we implement

bmckerry (approved):

sentry_streams/sentry_streams/adapters/arroyo/reduce.py
```diff
@@ -66,6 +76,65 @@ def accumulator(self, result: Any, value: BaseValue[RoutedValue]) -> RoutedValue
...
+        self.start_time: float | None = None
+        self.metrics = get_metrics()
+        self.tags = {"step": name}
+
+    def add(self, value: Message[Any]) -> Self:
+        if self.start_time is None:
+            self.start_time = time.time()
+        self.metrics.increment("streams.pipeline.input.messages", tags=self.tags)
+        self.metrics.increment(
+            "streams.pipeline.input.bytes", tags=self.tags, value=sys.getsizeof(value.payload)
```
untitaker, line 92: `getsizeof` feels a bit sketchy because the number doesn't mean anything for complex non-string types. i think it's probably better to check the type and get `len()`, and to not emit any metric if the payload isn't actually bytes

sentry_streams/src/metrics.rs
```diff
@@ -0,0 +1,58 @@
...
+
+/// A simple UDP sink for sending StatsD metrics
+struct UdpMetricSink {
+    socket: UdpSocket,
+    addr: String,
+}
+
+impl UdpMetricSink {
+    fn new(host: &str, port: u16) -> Result<Self, std::io::Error> {
+        let socket = UdpSocket::bind("0.0.0.0:0")?;
```
untitaker, line 13: IIRC this was a bit tricky to get right without any dataloss in prod. i suggest to not use sockets directly and either use statsdproxy (like snuba) or cadence. at least `nonblocking` is necessary: https://github.com/getsentry/statsdproxy/blob/709c9cd22a1f3fa97660895f5432a3c7fee2887e/src/middleware/upstream.rs#L27

sentry_streams/sentry_streams/runner.py
```diff
@@ -146,6 +177,24 @@ def signal_handler(sig: int, frame: Any) -> None:
     type=str,
     help="The segment id to run the pipeline for",
 )
+@click.option(
```
untitaker, line 180: i think the tricky bit is how we'd get runtime metrics and application metrics tagged with the same set of global tags.

maybe there's an API in the streaming platform to get a set of global tags that the applpication can then add to its own config.

right now we do it the other way around (which I think is wrong): sentry has a mechanism for setting global tags, and imposes it onto arroyo.

sentry_streams/sentry_streams/adapters/arroyo/rust_arroyo.py
```diff
@@ -204,11 +235,22 @@ def sink(self, step: Sink[Any], stream: Route) -> Route:
...
+                has_error = output_size = None
+                try:
+                    result = step.object_generator()
+                    output_size = sys.getsizeof(result)
+                    return result
+                except Exception as e:
+                    has_error = True
+                    raise e
+                finally:
+                    output_metrics(step.name, has_error is True, start_time, output_size)
```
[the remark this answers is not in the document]
evanh: I added a TODO to make this step actually work. I'll have to come up with a different strategy to capture the full metrics in a separate PR.

sentry_streams/sentry_streams/adapters/arroyo/rust_arroyo.py
```diff
@@ -64,6 +70,27 @@
 logger = logging.getLogger(__name__)
 
 
+def input_metrics(name: str, message_size: int | None) -> float:
+    metrics = get_metrics()
+    tags = {"step": name}
```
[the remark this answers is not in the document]
evanh: The pipeline name is set as a default tag when the runner is started, it's added to every metric automatically.

sentry_streams/sentry_streams/adapters/arroyo/reduce.py
```diff
@@ -66,6 +75,64 @@ def accumulator(self, result: Any, value: BaseValue[RoutedValue]) -> RoutedValue
         return routed
 
 
+class MetricsAccumulator(Accumulator[Message[Any], Any]):
```
fpacifici, line 78: hint: `MetricsReportingAccumulator`, `ObservableAccumulator` ? `MetricsAccumulator` seems an accumulator of metrics.

sentry_streams/sentry_streams/metrics/metrics.py
```diff
@@ -0,0 +1,236 @@
...
+def get_metrics() -> Metrics:
+    if _metrics_backend is None:
+        return _dummy_metrics_backend
+    return _metrics_backend
+
+
+def get_size(obj: Any) -> int | None:
+    if isinstance(obj, (str, bytes)):
+        return len(obj)
+    return None
```
fpacifici, line 236: What;s your goal here for the other message types? 

I'd recommend making `get_size` a method (or a property if you like it better) of the message class so each can implement it properly. That way we could also make messages rely on a header populated at the consumer level to track the size of the message. That would not impact anybody else.

If you plan to support other message types later, that's ok, in such case I would recommend adding a todo so we can track the missing work and still move the method to the message class now so we deal with the right interface.

sentry_streams/sentry_streams/adapters/arroyo/reduce.py
```diff
@@ -66,6 +75,64 @@ def accumulator(self, result: Any, value: BaseValue[RoutedValue]) -> RoutedValue
...
+        self.metrics.timing(Metric.DURATION, duration, tags=self.tags)
+        return result
+
+    def merge(self, other: Self) -> Self:
+        self.acc.merge(other.acc)
+
+        return self
+
+
+class MetricsReduce(PipelineReduce[MeasurementUnit, InputType, OutputType]):
```
fpacifici, line 115: Same as above `MetricsReportingReducer` or others? This does not reduce metrics but it is a Reducer that reports metrics

sentry_streams/sentry_streams/adapters/arroyo/reduce.py
fpacifici, line 78: Also why does this extend `Accumulator[Message[Any, Any]] rather than being generic itself? It seems we are loosing type safety here as the accumulator is expected to be `Accumulator[Any, Any]` and we cast it as `Accumulator[PipelineMessage[InputType], OutputType]]` below. This could be enforced statically. 
Was there any issue you could not address ?

sentry_streams/sentry_streams/adapters/arroyo/reduce.py
```diff
@@ -66,6 +75,64 @@ def accumulator(self, result: Any, value: BaseValue[RoutedValue]) -> RoutedValue
...
+        return self.reduce.group_by
+
+    @property
+    def windowing(self) -> Window[MeasurementUnit]:
+        return self.reduce.windowing
+
+    @property
+    def aggregate_fn(self) -> Callable[[], Accumulator[PipelineMessage[InputType], OutputType]]:
+        return cast(Callable[[], Accumulator[PipelineMessage[InputType], OutputType]], self.acc)
+
```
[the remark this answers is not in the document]
fpacifici: I think the AI is right. 
```
self.acc.add(value.payload)
```

Why are we unwrapping the payload out of the message when passing it to the inner accumulator, if the MetricsAccumulator is just a thin wrapper we should pass the message as it is.
IF I am right then this should be covered by unit tests.

sentry_streams/sentry_streams/adapters/arroyo/rust_arroyo.py
```diff
@@ -64,6 +70,27 @@
 logger = logging.getLogger(__name__)
 
 
+def input_metrics(name: str, message_size: int | None) -> float:
```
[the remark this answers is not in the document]
fpacifici: ok, we really need to prioritize that  Consumer Framework abstraction between arroyo and our streams code. It is making it really hard to think in terms of design and modules.

sentry_streams/sentry_streams/adapters/arroyo/rust_arroyo.py
```diff
@@ -211,11 +244,19 @@ def sink(self, step: Sink[Any], stream: Route) -> Route:
                 bucket = step.bucket
                 thread_count = 1
 
-            object_generator = step.object_generator
+            # TODO: This object_generator is just used to get the name of the object that is being written.
+            # Fix this to wrap the actual step instead of just the object_generator.
+            # This will at least capture the number of calls to the step, if nothing else.
+            def wrapped_generator() -> str:
+                start_time = input_metrics(step.name, None)
```
fpacifici, line 251: I'd juse remove the metrics now and just leave the TODO here. So we do not create confusion on what works

sentry_streams/sentry_streams/adapters/arroyo/rust_arroyo.py
```diff
@@ -243,6 +284,22 @@
...
+                result = step.resolved_function(msg)
+                output_size = get_size(result)
+                return result
+            except Exception as e:
+                has_error = str(e.__class__.__name__)
+                raise e
+            finally:
+                output_metrics(step.name, has_error, start_time, output_size)
+
+        step = replace(step, function=wrapped_function)
```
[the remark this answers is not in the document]
fpacifici: > Additionally, the wrapped_function captures step by reference, leading to infinite recursion when step.resolved_function is called after step has been reassigned to include the wrapped_function.

Damn this is right https://github.com/getsentry/streams/actions/runs/16917186777/job/47933976833?pr=184
Please see what we miss to catch this in unit test rather than the integration one.

sentry_streams/sentry_streams/metrics/metrics.py
```diff
@@ -0,0 +1,236 @@
...
+METRICS_FREQUENCY_SEC = 10
+
+
+class Metric(Enum):
+    INPUT_MESSAGES = "streams.pipeline.input.messages"
+    INPUT_BYTES = "streams.pipeline.input.bytes"
+    OUTPUT_MESSAGES = "streams.pipeline.output.messages"
+    OUTPUT_BYTES = "streams.pipeline.output.bytes"
+    DURATION = "streams.pipeline.duration"
+    ERRORS = "streams.pipeline.errors"
```
fpacifici, line 25: I'd add a description in a comment, so this populates the sphinx docs.

sentry_streams/sentry_streams/metrics/metrics.py
fpacifici, line 1: This module can benefit of some unit tests.

sentry_streams/sentry_streams/adapters/arroyo/reduce.py
fpacifici, line 78: Unit tests please also. You can use a Dummy metrics backend that tracks what we produce like in snuba.

sentry_streams/sentry_streams/adapters/arroyo/rust_arroyo.py
fpacifici, line 1: The changes we have here should be tested. It is right that The adapter is not easy to test because it does not create a consumer usable in tests but we can move out the logic you added and add tests.

sentry_streams/sentry_streams/runner.py
evanh: Oh I see. There is actually a place for global tags to also be defined, when the metrics backend is instantiated. Those tags are injected into every metric call. I will add a `add_global_tag` function so that those tags can be modified dynamically.

sentry_streams/sentry_streams/metrics/metrics.py
evanh: I tried the approach of using a `.size()` method but Reduce steps operate on arroyo messages not pipeline.Message. So I used this function instead. I'll add a TODO for the other work.

fpacifici (approved):
Thanks

sentry_streams/tests/pipeline/test_metrics.py
```diff
@@ -0,0 +1,389 @@
...
+    def test_metric_enum_values(self) -> None:
+        assert Metric.INPUT_MESSAGES.value == "streams.pipeline.input.messages"
+        assert Metric.INPUT_BYTES.value == "streams.pipeline.input.bytes"
+        assert Metric.OUTPUT_MESSAGES.value == "streams.pipeline.output.messages"
+        assert Metric.OUTPUT_BYTES.value == "streams.pipeline.output.bytes"
+        assert Metric.DURATION.value == "streams.pipeline.duration"
+        assert Metric.ERRORS.value == "streams.pipeline.errors"
+
+
+class TestDummyMetricsBackend:
```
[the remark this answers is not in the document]
evanh: Because that was what the AI decided to do when I asked it to generate unit tests.

PR #190, opened 2025-07-30
feat(watermarks): add watermark support to the ArroyoStrategyDelegate

bmckerry:
This PR:
- makes it so that the ArroyoStrategyDelegate hangs on to watermarks until the combined committable of all messages returned by the OutputRetriever has offsets greater than a given watermark's committable for all partitions
- updates reduce/multiprocess delegate tests to reflect this change
- renames the `RustMessage` enum to `KafkaMessage` to signify messages read from a kafka source
- makes a new enum named `RustMessage` signifying messages read from kafka, or watermarks sent by the pipeline


sentry_streams/sentry_streams/pipeline/message.py
```diff
@@ -14,11 +14,14 @@
...
-from sentry_streams.rust_streams import PyAnyMessage, RawMessage
+from sentry_streams.rust_streams import PyAnyMessage, PyWatermark, RawMessage
 
 TPayload = TypeVar("TPayload")
 
-RustMessage = Union[PyAnyMessage, RawMessage]
+# Represents the Rust message types which can be sent into a delegate step
+RustMessage = Union[PyAnyMessage, PyWatermark, RawMessage]
+# Represents the Rust message types which are actually processed by delegate steps
+KafkaMessage = Union[PyAnyMessage, RawMessage]
```
fpacifici, line 24: I'd recommend against this separation.  There are two issues.
- RustMessage represented the message that we can move back and forth between Rust and Python, it does not necessarily represent a KafkaMessage, so we should not rename it to KafkaMessage.
- If you need to extend RustMessage adding another type you should have a Union of a Union and the new type.

```
RustApplicationMessage = Union[PyAnyMessage, RawMessage]
RustMessage = Union[RustApplicationMessage, RawMessage]
```

sentry_streams/sentry_streams/adapters/arroyo/multi_process_delegate.py
```diff
@@ -45,7 +45,7 @@ def process_message(
 
 def mapped_msg_to_rust(
     message: ArroyoMessage[Union[FilteredPayload, Message[TMapOut]]],
-) -> Tuple[RustMessage, Committable] | None:
+) -> Tuple[KafkaMessage, Committable] | None:
```
fpacifici, line 48: Please consider doing the refactoring of the enum in a PR and put the application logic change in a separate PR. The wide change induced by the no brainer refactoring makes it hard to review the important part of your PR

sentry_streams/sentry_streams/adapters/arroyo/rust_step.py
```diff
@@ -273,23 +273,56 @@ class ArroyoStrategyDelegate(RustOperatorDelegate, Generic[TStrategyIn, TStrateg
...
+
+    def __yield_messages(self) -> Iterable[Tuple[RustMessage, Committable]]:
+        for message, committable in self.__retriever.fetch():
+            yield (message, committable)
+            self.__globbed_committable.update(committable)
+            watermarks = self.__watermarks
+            for wm in watermarks:
+                if self.__should_send_watermark(wm):
+                    yield (wm, wm.payload)
+                    self.__watermarks.remove(wm)
```
fpacifici, line 310: This logic will be quite delicate. It is very important to have it thoroughly tested. Now we are relying on the multiprocess delegate and reduce delegate tests to cover this. Such approach is fragile because those tests may assume the ArroyoStrategyDelegate works and not cover this logic. 

There is a `test_delegate.py`  file that tests the `RustOperatorDelegate` in a simpler environment than multiprocess and Reduce. Please add test cases there to validate that:
- The delegate sends through the right Watermark after seeing the committables
- The corner cases where partial committables are returned so we need to accumulate multiple of them before sending a watermark 
- Whether the watermark is send before or after the message in the right way.

sentry_streams/sentry_streams/pipeline/message.py
```diff
@@ -214,12 +217,14 @@ def __setstate__(self, state: Mapping[str, Any]) -> None:
...
-def rust_msg_equals(msg: RustMessage, other: RustMessage) -> bool:
+def pipeline_msg_equals(msg: KafkaMessage, other: KafkaMessage) -> bool:
     """
-    PyAnyMessage/RawMessage are exposed by Rust and do not have an __eq__ method
+    PyAnyMessage/PyWatermark/RawMessage are exposed by Rust and do not have an __eq__ method
     as of now. That would require delegating equality to python anyway
     as the payload is a PyAny
     """
+    if type(msg) is not type(other):
+        return False
```
fpacifici, line 227: Why is this needed ?

sentry_streams/sentry_streams/rust_streams.pyi
```diff
@@ -108,7 +108,10 @@ class RawMessage:
 class PyWatermark:
     def __init__(
         self,
-        committable: dict[Tuple[str, int], int],
+        payload: dict[tuple[str, int], int],
```
fpacifici, line 111: Why did this become `payload` ?

sentry_streams/tests/adapters/arroyo/test_reduce_delegate.py
```diff
@@ -34,19 +43,19 @@ def test_retriever() -> None:
     timestamp = datetime.now()
     retriever.submit(
         ArroyoMessage(
-            Value("payload", {Partition(Topic("test_topic"), 0): 100}, timestamp),
+            Value("payload", {Partition(Topic("topic0"), 0): 100}, timestamp),
```
fpacifici, line 46: why this change ?

sentry_streams/sentry_streams/adapters/arroyo/multi_process_delegate.py
bmckerry: Will do

sentry_streams/sentry_streams/rust_streams.pyi
bmckerry: I made this `payload` to keep it in line with the fields for `PyMessage`/`RawMessage`. This allows [assert_equal_batches()](https://github.com/getsentry/streams/blob/main/sentry_streams/tests/adapters/arroyo/test_reduce_delegate.py#L127-L135) to work without a bunch of annoying `cast()`s

sentry_streams/sentry_streams/adapters/arroyo/rust_step.py
```diff
@@ -279,17 +279,53 @@ def __init__(
...
+    def flush(self, timeout: float | None = None) -> Iterable[Tuple[PipelineMessage, Committable]]:
         self.__inner.join(timeout)
-        return self.__retriever.fetch()
+        return self.__yield_messages()
+
+    def watermarks(self) -> MutableSequence[PyWatermark]:
+        return self.__watermarks
+
+    def globbed_committable(self) -> Committable:
+        return self.__globbed_committable
```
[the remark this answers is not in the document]
bmckerry: These are exposed so I can validate they have the right structure in tests. I wasn't sure how else to do this

sentry_streams/tests/adapters/arroyo/test_delegate.py
```diff
@@ -60,3 +79,86 @@ def make_msg(payload: str) -> RustMessage:
...
+    # Message with the same committable as a watermark triggers the watermark
+    delegate.submit(*build_rust_msg("payload", 0, build_committable(1, 42)))
+    ret = list(delegate.poll())
+
+    expected = [
+        build_rust_msg("transformed payload", 0, build_committable(1, 42)),
+        build_watermark(
+            build_committable(1, 42),
+            timestamp=0,
+        ),
```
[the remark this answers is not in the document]
bmckerry: I've changed the test to be the following:
- message at offset 42 is received
- watermark for offset 42 is received
- watermark for offset 43 is received
- delegate is polled for results, only the message and watermark@42 should be sent

This can happen because the message at offset 43 could have been routed to a different branch, but its respective watermark would still go to the delegate because watermarks will be broadcast on all branches of a Router.

As for the ordering of watermarks vs messages, please refer to my comment above

fpacifici (approved):
Please see my last message before merging

PR #193, opened 2025-08-05
feat(batch): TumblingWindow by batch time

victoria-yining-huang:
https://linear.app/getsentry/issue/ENG-4503/add-the-ability-to-close-batches-by-time

tested the example `simple_batching.py` locally with the corresponding config file, behaving as expected. Tried:
- only running with batch_size
- only running with batch_timedelta
- running with batch_size and batch_timedelta, flush a batch by size is full first
- running with batch_size and batch_timedelta, flush a batch by batch time has passed first

### DIFF
--- a/sentry_streams/tests/test_pipeline.py
+++ b/sentry_streams/tests/test_pipeline.py
@@ -7,6 +7,7 @@ from arroyo.processing.strategies.abstract import ProcessingStrategy
 
 from sentry_streams.adapters.arroyo.reduce import (
     ArroyoAccumulator,
+    TimeWindowedReduce,
     build_arroyo_windowed_reduce,
 )
 from sentry_streams.adapters.arroyo.routes import Route
@@ -308,11 +309,12 @@ def test_batch_step_only_window_timedelta() -> None:
 
     reduce_window = TumblingWindow(window_timedelta=window_timedelta)
 
-    with mock.patch("sentry_streams.adapters.arroyo.reduce.TimeWindowedReduce") as MockReduce:
-        build_arroyo_windowed_reduce(reduce_window, acc, next_step, route)
-
-        args, _ = MockReduce.call_args
-        assert args[0] == args[1] == window_timedelta.total_seconds()
+    returned_reduce = build_arroyo_windowed_reduce(reduce_window, acc, next_step, route)
+    assert isinstance(returned_reduce, TimeWindowedReduce)
+    size = slide = window_timedelta.total_seconds()
+    assert returned_reduce.window_count == int(size / slide)
+    assert returned_reduce.window_size == int(size)
+    assert returned_reduce.window_slide == int(slide)
 
 
 def test_batch_step_only_window_size() -> None:


sentry_streams/sentry_streams/pipeline/pipeline.py
```diff
@@ -480,16 +487,22 @@ class Batch(
...
     # by either size or time
-    batch_size: MeasurementUnit
+
+    batch_size: int | None = None
+    batch_timedelta: timedelta | None = None
     step_type: StepType = StepType.REDUCE
 
+    def __post_init__(self) -> None:
+        if self.batch_size is None and self.batch_timedelta is None:
+            raise ValueError("Exactly one of batch_size or batch_timedelta must be set.")
```
john-z-yang, line 497:
```suggestion
            raise ValueError("At least one of batch_size or batch_timedelta must be set.")
```

sentry_streams/sentry_streams/adapters/arroyo/reduce.py
```diff
@@ -276,39 +275,48 @@ def build_arroyo_windowed_reduce(
...
-        case TumblingWindow(window_size):
-
+        case TumblingWindow(window_size, window_timedelta):
             arroyo_acc = ArroyoAccumulator(accumulator)
 
-            match window_size:
-                case int():
-                    return Reduce(
-                        window_size,
-                        float("inf"),
```
john-z-yang, line 287: I am very curious to why simple_batching.py did not work with this, because it seems like it is setting the batch size to be 3
```python
batch_size=3
```
If it is following this code path, it is set max_batch_time to infinity, and it should always wait for the batch size to reach 3 before closing it.

sentry_streams/sentry_streams/adapters/arroyo/reduce.py
```diff
@@ -276,39 +275,48 @@ def build_arroyo_windowed_reduce(
...
+                    cast(
+                        Callable[
+                            [FilteredPayload | TPayload, BaseValue[TPayload]],
+                            FilteredPayload | TPayload,
+                        ],
+                        arroyo_acc.accumulator,
+                    ),
+                    arroyo_acc.initial_value,
+                    msg_wrapper,
+                )
```
john-z-yang, line 319: Nit: I feel like it might be better to fold this branch into the first branch to avoid duplicate code.

Maybe something like
```python
if window_size is not None:
    return Reduce(
        window_size,
        window_timedelta or float("inf"),
        ...
    )
return TimeWindowedReduce(...)
```

sentry_streams/sentry_streams/pipeline/pipeline.py
```diff
@@ -466,6 +467,12 @@ def aggregate_fn(self) -> Callable[[], Accumulator[Message[InputType], OutputTyp
 BatchInput = TypeVar("BatchInput")
 
 
+@dataclass
+class WindowSize(Generic[MeasurementUnit, InputType]):
+    batch_timedelta: MeasurementUnit
+    batch_size: MeasurementUnit
```
john-z-yang, line 473: I am not familiar with python's type calculus, but I think this might imply that `batch_timedelta` and `batch_size` are of the same type. Which is not what we want.

Maybe
```python
@dataclass
class WindowSize(Generic[TimeUnit, SizeUnit, InputType]):
    batch_timedelta: TimeUnit
    batch_size: SizeUnit
```

But I feel like since you have already separated the 2, just using concrete types might be better here?

sentry_streams/sentry_streams/adapters/arroyo/reduce.py
victoria-yining-huang: i was having the same confusion yesterday. Anyway, today everything is working as expected

sentry_streams/sentry_streams/pipeline/pipeline.py
victoria-yining-huang: oops this class is actually not used anywhere. Deleted

john-z-yang (approved):
LGTM, good work!

PR #194, opened 2025-08-06
feat(rust): Provide a rust runner to start consumer in CLI in rust

john-z-yang:
Provides a way to let users to start their pipeline in rust.

Also added some argument parsing facilities that lets users to build CLI easily.

i.e.
```rust
use std::error::Error;

use rust_streams::run::{run, Args};

fn main() -> Result<(), Box<dyn Error>> {
    let args = Args::parse();

    run(args)
}
```

```bash
cargo run -- \
	--config-file ~/code/streams/sentry_streams/sentry_streams/deployment_config/simple_batching.yaml \
	--segment-id 0 \
	--adapter-name rust_arroyo \
	--exec-path ../../../../.venv/bin/python \
	--module-paths ../../../../.venv/bin/python \
	--module-paths ../../../../ \
	--module-paths ../../../../.venv/lib/python3.11/site-packages \
	--module-paths /opt/homebrew/Cellar/python@3.11/3.11.12_1/Frameworks/Python.framework/Versions/3.11/lib/python311.zip \
	--module-paths /opt/homebrew/Cellar/python@3.11/3.11.12_1/Frameworks/Python.framework/Versions/3.11/lib/python3.11 \
	--module-paths /opt/homebrew/Cellar/python@3.11/3.11.12_1/Frameworks/Python.framework/Versions/3.11/lib/python3.11/lib-dynload \
	~/code/streams/sentry_streams/sentry_streams/examples/simple_batching.py
```

They can also start the pipeline programmatically should they chose to
```rust
use std::error::Error;
use rust_streams::run::{run, Args};

fn main() -> Result<(), Box<dyn Error>> {
    run(Args {
        // settings here
    })
}
```

### DIFF
--- a/sentry_streams/sentry_streams/examples/rust_simple_map_filter/rust_transforms/Cargo.toml
+++ b/sentry_streams/sentry_streams/examples/rust_simple_map_filter/rust_transforms/Cargo.toml
@@ -7,6 +7,10 @@ edition = "2021"
 name = "metrics_rust_transforms"
 crate-type = ["cdylib"]
 
+[[bin]]
+name = "runner"
+path = "src/main.rs"
+
 [dependencies]
 pyo3 = { version = "0.24" }
 serde = { version = "1.0", features = ["derive"] }
@@ -20,3 +24,4 @@ serde_json = "1.0"
 [dependencies.rust_streams]
 path = "../../../../"  # Point to the main rust_streams crate
 default-features = false  # Disable default features to avoid loading pyo3 a second time
+features = ["cli"]
--- a/sentry_streams/sentry_streams/runner.py
+++ b/sentry_streams/sentry_streams/runner.py
@@ -1,7 +1,6 @@
 import importlib
 import json
 import logging
-import signal
 from typing import Any, Optional, cast
 
 import click
@@ -60,21 +59,22 @@ def iterate_edges(
                     step_streams[branch_name] = next_step_stream[branch_name]
 
 
-def runner(
+def load_runtime(
     name: str,
     log_level: str,
     adapter: str,
     config: str,
     segment_id: Optional[str],
     application: str,
-) -> None:
-    pipeline_globals: dict[str, Any] = {}
+) -> Any:
 
     logging.basicConfig(
         level=log_level,
         format="%(asctime)s - %(levelname)s - %(message)s",
         datefmt="%Y-%m-%d %H:%M:%S",
     )
+
+    pipeline_globals: dict[str, Any] = {}
 
     with open(application) as f:
         exec(f.read(), pipeline_globals)
@@ -119,14 +119,7 @@ def runner(
 
     iterate_edges(pipeline, translator)
 
-    def signal_handler(sig: int, frame: Any) -> None:
-        logger.info("Signal received, terminating the runner...")
-        runtime.shutdown()
-
-    signal.signal(signal.SIGINT, signal_handler)
-    signal.signal(signal.SIGTERM, signal_handler)
-
-    runtime.run()
+    return runtime
 
 
 @click.command()
@@ -184,14 +177,8 @@ def main(
     segment_id: Optional[str],
     application: str,
 ) -> None:
-    runner(
-        name,
-        log_level,
-        adapter,
-        config,
-        segment_id,
-        application,
-    )
+    runtime = load_runtime(name, log_level, adapter, config, segment_id, application)
+    runtime.run()
 
 
 if __name__ == "__main__":
--- a/sentry_streams/src/run.rs
+++ b/sentry_streams/src/run.rs
@@ -1,0 +1,100 @@
+use std::ffi::OsString;
+
+use clap::Parser;
+use pyo3::prelude::*;
+
+use crate::utils::traced_with_gil;
+
+#[derive(Parser, Debug)]
+pub struct Args {
+    #[command(flatten)]
+    pub py_config: PyConfig,
+
+    #[command(flatten)]
+    pub runtime_config: RuntimeConfig,
+}
+
+impl Args {
+    pub fn parse() -> Self {
+        Parser::parse()
+    }
+}
+
+#[derive(Parser, Debug)]
+pub struct PyConfig {
+    /// Path of the python interpreter executable
+    #[arg(short, long)]
+    pub exec_path: Option<OsString>,
+
+    #[arg(short, long)]
+    /// Paths to the relevant python modules
+    pub module_paths: Option<Vec<OsString>>,
+}
+
+#[derive(Parser, Debug)]
+pub struct RuntimeConfig {
+    /// The name of the Sentry Streams application
+    #[arg(short, long)]
+    pub name: String,
+
+    /// The name of the Sentry Streams application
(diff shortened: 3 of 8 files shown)


sentry_streams/src/run.rs
```diff
@@ -0,0 +1,55 @@
...
+            .getattr("load_physical_plan")?
+            .call1((
+                plan_config.adapter_name,
+                plan_config.config_file,
+                plan_config.segment_id,
+                plan_config.application_name,
+            ))?
+            .downcast::<PyList>()?
+            .get_item(0)?
+            .downcast::<ArroyoConsumer>()
```
john-z-yang, line 41: This doesn't work right now. I am going to try 2 more things defining some rust data structure that contains the mapping between str → rust functions and send that to ArroyoConsumer via python instead of trying to cast ArroyoConsumer in rust

sentry_streams/src/consumer.rs
```diff
@@ -59,6 +59,10 @@ pub struct ArroyoConsumer {
     concurrency_config: Arc<ConcurrencyConfig>,
 }
 
+impl ArroyoConsumer {
+    pub fn load_rust_native_fns(&mut self) {}
```
john-z-yang, line 63: My goal is to call this function in rust, so I can supply a str → rust function mapping that lets the consumer create native rust steps

sentry_streams/src/run.rs
john-z-yang: I am not sure why this is not working because I am using it pretty similar to the way that is described in [the documentation](https://pyo3.rs/main/doc/pyo3/types/trait.pyanymethods#tymethod.downcast).

I wonder if it's because it is created in python, i.e. with
```python
self.__consumers[source_name] = ArroyoConsumer(
    source=source_name,
    kafka_config=build_kafka_consumer_config(source_name, source_config),
    topic=step.stream_name,
    schema=step.stream_name,
)
```

instead of it being created it rust?

sentry_streams/sentry_streams/examples/rust_simple_map_filter/rust_transforms/Cargo.toml
fpacifici, line 1: Shouldn't this become a `bin` crate that imports stream as a crate rather than the other way around ?
In other words, do we still need to expose the cdylib ?

sentry_streams/sentry_streams/runner.py
```diff
@@ -55,22 +55,14 @@ def iterate_edges(
...
-def runner(
-    name: str,
-    log_level: str,
+def load_runtime(
     adapter: str,
     config: str,
     segment_id: Optional[str],
     application: str,
-) -> None:
+) -> Any:
```
fpacifici, line 63: Please feel free to add the right type that is an adapter. 
Most likely you will run into generic issues, then you can cast it to the type you need. It will still be better than Any

sentry_streams/sentry_streams/runner.py
```diff
@@ -158,7 +143,22 @@ def main(
...
+    runtime = load_runtime(adapter, config, segment_id, application)
+
+    def signal_handler(sig: int, frame: Any) -> None:
+        logger.info("Signal received, terminating the runner...")
+        runtime.shutdown()
+
+    signal.signal(signal.SIGINT, signal_handler)
+    signal.signal(signal.SIGTERM, signal_handler)
+
+    runtime.run()
```
fpacifici, line 161: Why does mypy like this? `runtime ` is Any

sentry_streams/src/run.rs
fpacifici, line 1: Do you plan to keep both the python runner and the rust runner or to remove the python runner (I like the idea of having only one but as long as the rust one is a thin wrapper, no big deal)

sentry_streams/src/run.rs
```diff
@@ -0,0 +1,99 @@
...
+        PyResult::Ok(runtime)
+    })?;
+
+    traced_with_gil!(|py| {
+        runtime
+            .bind(py)
+            .call_method0("run")
+            .expect("Unable to start runtime");
+    });
+
```
fpacifici, line 97: Is this the place where you had a hard time downcasting the consumer object provided by python ?

sentry_streams/src/run.rs
```diff
@@ -0,0 +1,99 @@
...
+    /// The segment id to run the pipeline for
+    #[arg(short, long)]
+    pub segment_id: Option<String>,
+
+    /// The name of the application
+    pub application_name: String,
+}
+
+pub fn run(args: Args) -> Result<(), Box<dyn std::error::Error>> {
+    pyo3::prepare_freethreaded_python();
```
fpacifici, line 47: I don't think we can use freethreaded python now. Assuming this starts python with freethreading. That's because, as I understand it, it requires the application to take care of race conditions more and our application does not do that.

sentry_streams/src/run.rs
```diff
@@ -0,0 +1,99 @@
...
+            ))?
+            .unbind();
+
+        {
+            let runtime = runtime.clone_ref(py);
+            ctrlc::set_handler(move || {
+                traced_with_gil!(|py| {
+                    runtime
+                        .bind(py)
+                        .call_method0("shutdown")
```
fpacifici, line 82: I don't think this will work.

```
    def shutdown(self) -> None:
        """
        Shutdown the arroyo processors allowing them to terminate the inflight
        work.
        """
        raise NotImplementedError
```

While the signal management code remained in the python version the signal handler that actually takes care of terminating the consumer is in `consumer.rs`

fpacifici (commented):
Thanks, this is good. 

A few high level questions:
1. Do you plan to go a step further and introduce some python pipeline primitives that reference rust function by name which are statically linked ? Or is that out of scope ?
2. Do you plan to have a rust and a python cli or to keep only the rust one ?

sentry_streams/sentry_streams/examples/rust_simple_map_filter/rust_transforms/Cargo.toml
john-z-yang: Maybe this comment is going onto the wrong file, but `sentry_streams/sentry_streams/examples/rust_simple_map_filter/rust_transforms/Cargo.toml` does have a `bin` right now.

This is in the same example where Markus had his rust function example, I will refactor it into a separate example.

sentry_streams/src/run.rs
john-z-yang: Yes, I tried a couple of things and they all did not work. Markus and I thought to reduce the scope down a little bit to just have a way to start things using rust first, and worry about hydrating the runtime with pure rust functions later.

sentry_streams/src/run.rs
john-z-yang: I think we can remove the python runner if we like, and just expose the rust run function to python with pyo3

sentry_streams/src/run.rs
fpacifici: IT works because rust also register the signal handler and python signal handler is never reached as the rust code intercepts it. 
I forgot to remove the python code when I added the rust implementation. I had tried to do it inpython first and could not figure it out. It worked fine in rust

PR #195, opened 2025-08-11
feat: add integration test for WatermarkEmitter

bmckerry:
This PR adds a test for WatermarkEmitter that ensures the WatermarkCommitOffsets step sends a CommitRequest when expected.

It does this by controlling when WatermarkEmitter sends watermarks via conditional compilation of the `current_epoch()` function. rust_analyzer doesn't place nice with conditional compilation, so I had to add some ignores to our linting. See below comments for details.

### DIFF
--- a/sentry_streams/src/watermark.rs
+++ b/sentry_streams/src/watermark.rs
@@ -117,15 +117,18 @@ impl ProcessingStrategy<RoutedValue> for WatermarkEmitter {
 mod tests {
     use super::*;
     use crate::commit_policy::WatermarkCommitOffsets;
+    use crate::consumer::build_chain;
     use crate::fake_strategy::{assert_watermarks_match, FakeStrategy};
     use crate::messages::Watermark;
     use crate::mocks::set_timestamp;
+    use crate::operators::RuntimeOperator;
     use crate::routes::Route;
-    use crate::testutils::{build_routed_value, make_committable, make_lambda};
-    use crate::transformer::build_map;
+    use crate::testutils::{build_routed_value, make_committable, make_lambda, make_msg};
     use crate::utils::traced_with_gil;
     use pyo3::ffi::c_str;
+    use pyo3::prelude::*;
     use pyo3::IntoPyObjectExt;
+    use sentry_arroyo::processing::strategies::run_task_in_threads::ConcurrencyConfig;
     use sentry_arroyo::processing::strategies::ProcessingStrategy;
     use sentry_arroyo::types::Topic;
     use std::collections::HashMap;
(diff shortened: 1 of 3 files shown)


sentry_streams/src/mocks.rs
```diff
@@ -0,0 +1,18 @@
+use std::cell::Cell;
+
+thread_local! {
+    static TIMESTAMP: Cell<u64> = const { Cell::new(0) };
+}
+
+// Needed as rust_analyzer doesn't play nice with conditional compilation
+#[allow(dead_code)]
```
bmckerry, line 8: Since the functions in `mocks.rs` are only called when the `test` feature is enabled, and `time_helpers.rs` are only called when `tests` isn't enabled, rust_analyzer considers these to all be dead code (despite the fact that one of these will always be true).

sentry_streams/src/watermark.rs
```diff
@@ -177,11 +181,71 @@ mod tests {
 
             // submitted WatermarkMessage contains the last seen committable
             watermark.last_sent_timestamp = 0;
+            set_timestamp(20);
```
bmckerry, line 184: needed as we're using the mock `current_epoch()` here now

sentry_streams/src/watermark.rs
```diff
@@ -5,16 +5,13 @@ use sentry_arroyo::processing::strategies::{
...
 
-/// Returns the current Unix epoch
-fn current_epoch() -> u64 {
-    SystemTime::now()
-        .duration_since(UNIX_EPOCH)
-        .unwrap()
-        .as_secs()
-}
+#[cfg(test)]
+use crate::mocks::current_epoch;
```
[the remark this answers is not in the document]
bmckerry: Unfortunately it doesn't override the import if I do that, the `mock` module just isn't used

Also moving the `use crate::mocks::current_epoch;` into the test import list makes rust_analyzer throw a bunch of red squiggles because it always runs with the `test` feature enabled (same reason the `#[cfg(not(test))]` bit looks greyed out if you open it)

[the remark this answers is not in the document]
evanh: Got it, just wanted to check.

sentry_streams/src/watermark.rs
```diff
@@ -177,11 +180,71 @@ mod tests {
...
         })
     }
+
+    #[test]
+    fn test_watermark_to_commit_step() {
+        // TODO: once Router works with watermarks, ensure commit request is not returned
+        // from commit step if messages from all routes haven't reached commit step
+        crate::testutils::initialize_python();
+        traced_with_gil!(|py| {
+            let callable = make_lambda(py, c_str!("lambda x: x"));
```
fpacifici, line 199: In order to make this an actual integration test, please use the `build_chain` method to build the strategy https://github.com/getsentry/streams/blob/main/sentry_streams/src/consumer.rs#L184-L192.

By building the strategy in this test manually you are indeed testing that if you wire a watermarkemitter to a commitoffsets, watermarks are managed properly, though you are not testing that the consumer assembles the strategy correctly. That was the goal of the integration test: build a consumer in a realistic way and send message through.

Now, using `build_chain` skill skips a couple of steps (the real consumer management, and the `add_step` of the consumer struct`). While desirable to test them they are much harder to pull off. 
Though the `build_chain` part is the most relevant one as it is the one that ensures the watermark emitter and commit strategy are wired up. and it is the one that builds the strategy itself like it would do in prod.

If somebody refactored the `build_chain` code in a wrong way this integration test would not catch it, thus it would not be an integration test.

PR #198, opened 2025-08-13
fix(gcssink): make it compile with unused variable thread_count

victoria-yining-huang:


fpacifici (commented):
Please add a unit test that instantiates the GCSSink operator from Python. THe example is not a test that would run in CI, so not going to prevent the issue.
You just need to create a unit test in python that does one of these:
1. just instantiate each of `RuntimeOperator` like `RuntimeOperator.GCSSink(route, bucket, object_generator, thread_count)`. That is enough to validate the interfaces are coherent between the languages 
2. or build a consumer and use add_step to add the operators.

fpacifici (approved):

PR #200, opened 2025-08-27
fix(python_adapter): add test for ArroyoDelegate in rust

bmckerry:
Followup to https://github.com/getsentry/streams/pull/202

This PR adds a test for using the ArroyoDelegate from the PythonAdapter

### DIFF
--- a/sentry_streams/src/messages.rs
+++ b/sentry_streams/src/messages.rs
@@ -453,12 +453,13 @@ impl From<Py<PyAny>> for PyStreamingMessage {
     fn from(value: Py<PyAny>) -> Self {
         traced_with_gil!(|py| {
             let bound = value.clone_ref(py).into_bound(py);
-
-            if let Ok(content) = bound.downcast::<PyAnyMessage>() {
+            if bound.is_instance_of::<PyAnyMessage>() {
+                let content = bound.downcast::<PyAnyMessage>()?;
                 Ok(PyStreamingMessage::PyAnyMessage {
                     content: content.clone().unbind(),
                 })
-            } else if let Ok(content) = bound.downcast::<RawMessage>() {
+            } else if bound.is_instance_of::<RawMessage>() {
+                let content = bound.downcast::<RawMessage>()?;
                 Ok(PyStreamingMessage::RawMessage {
                     content: content.clone().unbind(),
                 })


sentry_streams/sentry_streams/adapters/arroyo/test_delegate.py
```diff
@@ -0,0 +1,62 @@
...
+def arroyo_msg_to_rust(
+    message: ArroyoMessage[Union[FilteredPayload, RustMessage]],
+) -> Tuple[RustMessage, Committable] | None:
+    if isinstance(message.payload, FilteredPayload):
+        return None
+    committable = {
+        (partition.topic.name, partition.index): offset
+        for partition, offset in message.committable.items()
+    }
+    return (message.payload, committable)
```
fpacifici, line 43: These seem unnecessary.
They seem 95% copies of these https://github.com/getsentry/streams/blob/main/sentry_streams/sentry_streams/adapters/arroyo/multi_process_delegate.py#L46-L91

The difference seems toi be that your `RunTask` op takes RustMessage and return RustMessage rather than taking and returning a PyAnyMessage.
If you make your function take and return a PyAnyMessage you should be able to avoid these methods entirely and actually test the real logic we run in production instead, which would be recommended.

sentry_streams/src/python_operator.rs
```diff
@@ -545,4 +546,33 @@ class RustOperatorDelegateFactory:
...
+                instance.unbind(),
+                Box::new(Noop {}),
+            );
+
+            let message = make_msg(py, "ok");
+            let res = operator.submit(message);
+            assert!(res.is_ok());
+
+            let commit_request = operator.poll();
+            assert!(commit_request.is_ok());
```
fpacifici, line 575: How is this returning OK now that your fix has not been merged yet ?

fpacifici (changes requested):

sentry_streams/sentry_streams/adapters/arroyo/test_delegate.py
bmckerry: I tried originally doing what you're suggesting, but spent >30 min wrestling with mypy to make it happy regarding the `TMapIn`/`TMapOut` generics in the ArroyoStrategyDelegate.

I can spend more time on this to get the typing right, it was just a nightmare.

PR #202, opened 2025-08-28
fix(python_adapter): more __yield_messages fixes

bmckerry:
https://github.com/getsentry/streams/pull/203 fixed the initial error by returning messages as a Sequence instead of a generator, but this caused another issue where watermarks weren't handled by the python adapter's `handle_py_return_value` function.

This required implementing `From Py<PyAny> for Watermark` to allow converting the python `PyWatermark` returned from the `PythonAdapter` into a rust `PyWatermark`.

A rust test for this code path is coming in a future PR.

### DIFF
--- a/sentry_streams/src/python_operator.rs
+++ b/sentry_streams/src/python_operator.rs
@@ -278,8 +278,8 @@ mod tests {
     use super::*;
     use crate::fake_strategy::assert_messages_match;
     use crate::fake_strategy::FakeStrategy;
-    use crate::messages::make_py_int;
-    use crate::messages::{PyWatermark, WatermarkMessage};
+    use crate::messages::Watermark;
+    use crate::messages::WatermarkMessage;
     use crate::testutils::build_routed_value;
     use crate::testutils::make_committable;
     use pyo3::ffi::c_str;
@@ -300,6 +300,10 @@ class RustOperatorDelegate:
 
     def submit(self, payload, committable):
         self.committable = committable
+        # Handle watermark messages (PyWatermark objects)
+        if hasattr(payload, "committable"):
+            self.payload = payload
+            return
         if payload.payload == "ok":
             self.payload = payload
             return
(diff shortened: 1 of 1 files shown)


sentry_streams/src/messages.rs
```diff
@@ -473,6 +474,38 @@ impl From<Py<PyAny>> for PyStreamingMessage {
...
+            } else {
+                Err(pyo3::exceptions::PyTypeError::new_err(format!(
+                    "Message type is invalid: expected PyWatermark, got {}",
+                    bound.get_type().name().unwrap()
+                )))
+            }
+        })
+        .unwrap()
+    }
+}
```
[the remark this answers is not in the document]
fpacifici: Cursor is right. From cannot fail. You should use TryFrom for type conversions that can fail 
https://doc.rust-lang.org/std/convert/trait.TryFrom.html

fpacifici (changes requested):
Please see the issue with From

sentry_streams/src/python_operator.rs
```diff
@@ -73,15 +73,26 @@ impl PythonAdapter {
...
-            let message = Message::new_any_message(
-                RoutedValue {
-                    route: self.route.clone(),
-                    payload: RoutedValuePayload::PyStreamingMessage(payload.into()),
-                },
-                convert_py_committable(py, committable_dict).unwrap(),
-            );
-
-            self.transformed_messages.push_back(message);
+            let bound = payload.clone_ref(py).into_bound(py);
```
fpacifici, line 76: We discussed about doing a proper integration test as a followup but this is now changing the rust logic and this logic should be covered.

There are two options:
1. Add a test that specifically calls this method 
2. Amend one of the tests that rely on this copy of the Delegate to return a PyWatermark as well. https://github.com/getsentry/streams/blob/main/sentry_streams/src/python_operator.rs#L285-L316

sentry_streams/src/python_operator.rs
```diff
@@ -393,46 +405,10 @@ class RustOperatorDelegateFactory:
                     }
                 ))
             ));
-        })
-    }
-
-    #[test]
-    fn test_submit_watermark() {
```
bmckerry, line 400: This test was almost entirely redundant and only passed because `make_test_watermark` had the wrong route. I moved submitting a watermark into the above `test_submit_with_matching_route` test (the diff makes this look weird)

sentry_streams/src/python_operator.rs
```diff
@@ -340,20 +355,17 @@ class RustOperatorDelegateFactory:
         Message::new_any_message(routed_value, committable)
     }
 
-    fn make_test_watermark(py: Python<'_>) -> Message<RoutedValue> {
+    fn make_test_watermark() -> Message<RoutedValue> {
```
bmckerry, line 358: This was completely broken and only worked because it returned a message with the wrong route.

victoria-yining-huang (approved):

PR #203, opened 2025-09-02
fix(watermark): Return a list instead of a generator

evanh:
This fixes a bug in the code that expects a list instead of a Generator.


evanh:
> so why did the code typecheck if the caller expected a list and only got `Iterable`

I believe the caller is in Rust using pyo3. @bmckerry or @victoria-yining-huang could clarify specifically where this happens.

victoria-yining-huang (approved):
yes this works just pulled the branch and tested

untitaker:
@evanh i thikn the signature of flush is also wrong

PR #288, opened 2026-03-30
feat(streams): Add backpressure metrics for consumer strategies

fpacifici:
Rust arroyo does not record metrics for backpressure unless the exception is propagated back to the consumer.
This PR adds metrics to record backpressure events no matter where they happen together with a metrics that tracks how long a step is exerting backpressure for.

This is the first step to improve the throughput metrics.

1. Refactor how metrics are initialized. The initializer we have is for arroyo metrics. We were not recording anything from the rust code. Added unit tests as well.
2. Added a BackpressureTracker that keeps track of the backpressure state of every step and wraps an Arroyo strategy to intercept backrpessure events.
3. Allow the python code to specify the name of a step to the rust code (used to populate tags).

What would come next:
- buffered metrics to impact performance less
- success rate to compute the rate between success and backpressure
- more metrics recorders (log based to record periodically stats)


sentry_streams/src/operators.rs
```diff
@@ -91,13 +92,49 @@ pub enum RuntimeOperator {
...
 
+/// Stable label for `step` on backpressure metrics (one series per runtime operator).
+pub fn metric_label_for_runtime_operator(op: &RuntimeOperator) -> String {
+    fn route_seg(r: &Route) -> String {
+        if r.waypoints.is_empty() {
+            r.source.clone()
+        } else {
+            format!("{}>{}", r.source, r.waypoints.join(">"))
+        }
+    }
```
fpacifici: We cannot use the waypoints of the route to identify the step.
Steps have names in the pipeline. See an example here: https://github.com/getsentry/streams/blob/main/sentry_streams/sentry_streams/examples/simple_map_filter.py#L16-L20.
The adapter has access to the names: https://github.com/getsentry/streams/blob/main/sentry_streams/sentry_streams/pipeline/pipeline.py#L198. 

Now, when rust_arroyo.py creates the consumer by appending steps. We want to pass the names in order to manage assign the step name to the metrics.

PR #290, opened 2026-04-01
Add fast headers filter

fpacifici:
I tested the headers filter used in the SBC implementation. 
It seems that filter is terribly inefficient. This seemingly is due to the python/rust 
interaction when providing the headers to the python code.
We seem to be spending most of the time here https://github.com/getsentry/streams/blob/main/sentry_streams/src/messages.rs#L47-L82

I want to try SBC with a native rust implementation of the header filter.
This is going to be used here https://github.com/getsentry/super-big-consumers/blob/master/consumer/streaming_platform/pipelines/items_span.py#L28 
where the filtering logic is trivial (check that the value of a header equals an integer).

The goal is to see if the spans pipeline does not have that bottleneck anymore


sentry_streams/sentry_streams/pipeline/pipeline.py
```diff
@@ -476,6 +476,26 @@ def resolved_function(self) -> Callable[[Message[TIn]], bool]:
         return function_callable
 
 
+@dataclass
+class HeaderIntFilter(Transform[TIn, TIn], Generic[TIn]):
```
fpacifici: Please see the refactoring explained here https://github.com/getsentry/streams/pull/290/changes#r3022430003

evanh (approved):
Seems like a fair test to me.

PR #291, opened 2026-04-08
fix(kafka config): connect override_params from yaml config to rust arroyo

victoria-yining-huang:
also added tests to explicit assert on each field to make sure all fields are passed through to rust-arroyo


fpacifici (approved):
Please address the comments before merging

PR #299, opened 2026-04-22
Reduce the number of payload copy between rust and python

fpacifici:
I ran a number of load tests to understand where the bottlenecks of `streams` were.
It turns out that a major component is the repeated copy of payload and headers between rust memory and python memory. 

The current python message contains a number of properties like payload and headers. 
Each of them delegates the execution to the rust object.
Problem: referencing the `payload` attribute on the python message means calling a rust function that takes the GIL and copies the content of the payload from rust memory into python memory. This happens every time python code references the payload. Which is tons of times when processing the message.
Worse: when we reference headers, the headers are unpacked from librdkafka, parsed and made into a python sequence taking the GIL. Every single time. 

Adding a caching layer inside the message so that we copy payload, headers, timestamp and schema only once.
This should already show an improvement. Then I will try to exclude the headers from the picture entirely.

### DIFF
--- a/sentry_streams/sentry_streams/pipeline/message.py
+++ b/sentry_streams/sentry_streams/pipeline/message.py
@@ -11,7 +11,6 @@ from typing import (
     Tuple,
     TypeVar,
     Union,
-    cast,
 )
 
 from sentry_streams.rust_streams import PyAnyMessage, PyWatermark, RawMessage
@@ -96,6 +95,11 @@ class PyMessage(Generic[TPayload], Message[TPayload]):
     By making this type generic we can still get a lot of guarantees
     by the type checker when wiring python primitives together as
     it can ensure primitives are compatible with each other.
+
+    The Rust ``PyAnyMessage`` is built on first :meth:`to_inner` and then cached,
+    so normal use of the Python wrapper avoids pyo3 allocation until a step
+    needs the inner type. Pickle serializes only the Python fields; the cache is
+    cleared on unpickle so the inner is recreated when needed.
     """
 
     def __init__(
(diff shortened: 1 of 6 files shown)


mwarkentin:
Do you have any data from your tests that you can link here? Would be good to capture the artifact.

fpacifici:
> Do you have any data from your tests that you can link here? Would be good to capture the artifact.

I still have to compile data from the sandbox tests. 
But:
Screenshot 2026-04-22 at 2 22 44 PM

This represents the throughput of a test case I ran in a sandbox environment.
The runs per second of arroyo is proportional to the messages per second we process as there is no async processing in the test case.
The high peaks are those without multiple payload copies. The low peaks are those where I executed all the current logic

This is where the test code is https://github.com/getsentry/streams/compare/main...fpacifici/profile_filter

sentry_streams/sentry_streams/pipeline/message.py
```diff
@@ -105,44 +109,56 @@ def __init__(
...
     def size(self) -> int | None:
-        if isinstance(self.inner.payload, (str, bytes)):
-            return len(self.inner.payload)
+        if isinstance(self._payload, (str, bytes)):
+            return len(self._payload)
         return None
 
     def __repr__(self) -> str:
-        return f"PyMessage({self.inner.__repr__()})"
+        return (
```
evanh, line 140: Is there a reason not to call `to_inner` and then repr that inner object directly?

evanh (approved):

sentry_streams/sentry_streams/pipeline/message.py
markstory: Wouldn't that trigger the expensive allocation?

markstory (approved):
Makes sense to me.

sentry_streams/sentry_streams/pipeline/message.py
fpacifici: Yes, that's the reason. Instantiating the rust object causes memory copy, that is what caused the issue.

PR #307, opened 2026-04-27
feat(sentry-streams): Add native Rust batch step for PyAnyMessage

fpacifici:
Add a Rust `Batch` + `BatchStep` implementation for pipeline batching: accumulate `Py<PyAnyMessage>` per route, flush on size or time, emit a single `PyAnyMessage` whose payload is a list of row payloads, then drive buffered and synthetic watermarks and existing backpressure behavior.

**Behavior**
- `RuntimeOperator::Batch` builds `build_batch_step` in Rust (wired from `rust_arroyo` / `reduce()` when the batch config is native).
- Only `PyStreamingMessage::PyAnyMessage` rows are accepted; `RawMessage` is rejected with `InvalidMessage` for broker messages so offsets surface correctly.
- The GIL is used when materializing the batched list on flush, not on every append (rows are stored as `Py<PyAnyMessage>` after `Message::into_payload`).

**Tests**
- `tests::batch`: `Batch` in isolation (list output, committable merge, `should_flush`).
- `tests::step`: `BatchStep` as `ProcessingStrategy` (route forward, flush count, raw rejection).

Python stubs and adapter glue updated for the Batch operator path where applicable.

### DIFF
--- a/sentry_streams/src/batch_step.rs
+++ b/sentry_streams/src/batch_step.rs
@@ -35,7 +35,7 @@ fn invalid_message_submit_error(message: &Message<RoutedValue>) -> SubmitError<R
 }
 
 /// Count- and/or time-based window of `PyAnyMessage` elements for one route. Start with
-/// [`Batch::from_initial_message`]; add more with [`Batch::append`].
+/// [`Batch::from_initial`]; add more with [`Batch::append`].
 pub(crate) struct Batch {
     route: Route,
     max_batch_size: Option<usize>,
@@ -46,55 +46,40 @@ pub(crate) struct Batch {
 }
 
 impl Batch {
-    /// First element in a window: applies this message’s committable, consumes it, and sets the
-    /// time-bounded deadline when `max_batch_time` is set.
-    pub fn from_initial_message(
+    /// First element in a window. `committable` and `content` are taken from the same
+    /// [`RoutedValue`] (see [`BatchStep::submit`]); `Batch` only ever holds `PyAnyMessage` elements.
+    pub fn from_initial(
         route: Route,
         max_batch_size: Option<usize>,
         max_batch_time: Option<Duration>,
-        message: Message<RoutedValue>,
+        committable: BTreeMap<Partition, u64>,
+        first_element: Py<PyAnyMessage>,
     ) -> Self {
         let mut batch_offsets: BTreeMap<Partition, u64> = BTreeMap::new();
-        for (p, o) in message.committable() {
+        for (p, o) in committable {
             batch_offsets
                 .entry(p)
                 .and_modify(|e| *e = (*e).max(o))
                 .or_insert(o);
         }
-        let inner = message.into_payload();
-        let pysm = match inner.payload {
-            RoutedValuePayload::PyStreamingMessage(m) => m,
-            _ => unreachable!(),
-        };
-        let PyStreamingMessage::PyAnyMessage { content } = pysm else {
-            unreachable!();
-        };
         let batch_deadline = max_batch_time.map(Deadline::new);
         Self {
             route,
             max_batch_size,
             batch_deadline,
-            elements: vec![content],
+            elements: vec![first_element],
             batch_offsets,
         }
     }
 
-    /// Merges committable from the message, consumes it, and appends the `PyAnyMessage` element.
-    pub fn append(&mut self, message: Message<RoutedValue>) {
-        for (p, o) in message.committable() {
+    /// Merges this message’s `committable` into the running offset map and appends `content`.
+    pub fn append(&mut self, committable: BTreeMap<Partition, u64>, content: Py<PyAnyMessage>) {
+        for (p, o) in committable {
             self.batch_offsets
                 .entry(p)
                 .and_modify(|e| *e = (*e).max(o))
                 .or_insert(o);
         }
-        let inner = message.into_payload();
-        let pysm = match inner.payload {
-            RoutedValuePayload::PyStreamingMessage(m) => m,
-            _ => unreachable!(),
-        };
-        let PyStreamingMessage::PyAnyMessage { content } = pysm else {
-            unreachable!();
-        };
         self.elements.push(content);
     }
 
@@ -128,8 +113,9 @@ impl Batch {
         self.batch_offsets.clone()
     }
 
-    /// GIL: build list payload from stored `Py<PyAnyMessage>` elements, wrap in a batched `Message`.
-    pub fn build_stacked_message(&self) -> Result<Message<RoutedValue>, StrategyError> {
+    /// GIL: build list payload from stored `Py<PyAnyMessage>` elements, wrap in a batched
+    /// [`Message`].
+    pub fn flush(&self) -> Result<Message<RoutedValue>, StrategyError> {
         if self.elements.is_empty() {
             return Err(StrategyError::Other("Batch: empty window".into()));
         }
@@ -175,16 +161,15 @@ pub struct BatchStep {
     max_batch_time: Option<Duration>,
     /// `None` until the first `PyAnyMessage` in a window; then the open batch.
     batch: Option<Batch>,
-    /// Watermarks received while the current batch window is open; moved to
-    /// [`Self::pending_watermarks`] when the batch is emitted successfully.
+    /// Watermarks received while the current batch window is open; on successful batch send they
+    /// are appended to [`Self::outbound`].
     watermark_buffer: Vec<Message<RoutedValue>>,
-    /// Watermarks (and the synthetic batch watermark) queued for the next step; appended on
-    /// batch flush, drained on each [`ProcessingStrategy::poll`] (same deque + re-queue pattern as
-    /// `PythonAdapter`'s transformed message queue).
-    pending_watermarks: VecDeque<Message<RoutedValue>>,
+    /// Batched follow-ups and downstream backpressure: anything not yet accepted by
+    /// `next_step` (watermarks, synthetic batch watermark, or other rejects), drained on
+    /// [`Self::drain_outbound`].
+    outbound: VecDeque<Message<RoutedValue>>,
 
     message_carried_over: Option<Message<RoutedValue>>,
-    pending_downstream: VecDeque<Message<RoutedValue>>,
     commit_request_carried_over: Option<CommitRequest>,
 }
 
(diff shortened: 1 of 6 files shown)


sentry_streams/src/batch_step.rs
```diff
@@ -0,0 +1,665 @@
+//! Batches `PyAnyMessage` rows on a route, then [`BatchStep`] forwards one `PyAnyMessage` with a
+//! list payload and manages watermarks and backpressure to the next Arroyo strategy.
+//!
+//! Only `PyAnyMessage` streaming input is supported; `RawMessage` is rejected with
+//! `SubmitError::InvalidMessage` when the input is a broker message.
```
fpacifici, line 5: Add a TODO calling out that we are going to support this as well.

sentry_streams/src/batch_step.rs
```diff
@@ -0,0 +1,665 @@
...
+        InnerMessage::BrokerMessage(broker_message) => {
+            SubmitError::InvalidMessage(broker_message.into())
+        }
+        InnerMessage::AnyMessage(..) => {
+            panic!("BatchStep: invalid message on AnyMessage; Arroyo cannot surface InvalidMessage")
+        }
+    }
+}
+
+// --- Batch: accumulate `Py<PyAnyMessage>`; GIL on flush when building the list payload ---
```
fpacifici: Remove this comment

sentry_streams/src/batch_step.rs
```diff
@@ -0,0 +1,665 @@
...
+
+/// Count- and/or time-based window of `PyAnyMessage` rows for one route. The first row is added
+/// via [`Batch::from_first_row`]; more rows with [`Batch::append_row`].
+pub(crate) struct Batch {
+    route: Route,
+    max_batch_size: Option<usize>,
+    max_batch_time: Option<Duration>,
+    rows: Vec<Py<PyAnyMessage>>,
```
fpacifici: Rename `rows` as `elements` . Also adjust all other conventions and comments to talk about elements rather than rows.

sentry_streams/src/batch_step.rs
```diff
@@ -0,0 +1,665 @@
...
+    batch_deadline: Option<Deadline>,
+    batch_offsets: BTreeMap<Partition, u64>,
+}
+
+impl Batch {
+    pub fn from_first_row(
+        route: Route,
+        max_batch_size: Option<usize>,
+        max_batch_time: Option<Duration>,
```
fpacifici, line 54: We should not need to keep max_batch_time as part of the state of the struct. batch_deadline should be enough.

sentry_streams/src/batch_step.rs
```diff
@@ -0,0 +1,665 @@
...
+    }
+
+    pub fn apply_committable_from(&mut self, message: &Message<RoutedValue>) {
+        for (p, o) in message.committable() {
+            self.batch_offsets
+                .entry(p)
+                .and_modify(|e| *e = (*e).max(o))
+                .or_insert(o);
+        }
+    }
```
fpacifici: Please let's avoid having two methods, one to add an element and one for applying the committable. Let's only have an append method that takes message and committable and also applies the committable

sentry_streams/src/batch_step.rs
```diff
@@ -0,0 +1,665 @@
...
+            commit_request_carried_over: None,
+        }
+    }
+
+    fn is_quiesced(&self) -> bool {
+        self.message_carried_over.is_none()
+            && self.batch.as_ref().map_or(true, |b| b.is_empty())
+            && self.pending_downstream.is_empty()
+            && self.watermark_buffer.is_empty()
+    }
```
fpacifici: This is used only in join. Please move it there

sentry_streams/src/batch_step.rs
```diff
@@ -0,0 +1,665 @@
...
+
+    fn drain_pending_downstream(&mut self) -> Result<(), StrategyError> {
+        while let Some(msg) = self.pending_downstream.pop_front() {
+            if !self.submit_with_poll(msg)? {
+                break;
+            }
+        }
+        Ok(())
+    }
```
fpacifici: Let's simplify this for now. 
If we get a MessageRejected when trying to push the watermarks after sending the batch through, le'ts just abandon those watermarks and adda. TODO to address this

sentry_streams/src/batch_step.rs
```diff
@@ -0,0 +1,665 @@
...
+
+    fn submit(&mut self, message: Message<RoutedValue>) -> Result<(), SubmitError<RoutedValue>> {
+        if self.route != message.payload().route {
+            return self.next_step.submit(message);
+        }
+
+        match &message.payload().payload {
+            RoutedValuePayload::WatermarkMessage(_) => {
+                self.watermark_buffer.push(message);
```
fpacifici: If the batch is empty this must go through immediately, Please also add a test for this

sentry_streams/src/batch_step.rs
```diff
@@ -0,0 +1,665 @@
...
+        Ok(merge_commit_request(
+            self.commit_request_carried_over.take(),
+            c,
+        ))
+    }
+
+    fn submit(&mut self, message: Message<RoutedValue>) -> Result<(), SubmitError<RoutedValue>> {
+        if self.route != message.payload().route {
+            return self.next_step.submit(message);
+        }
```
fpacifici, line 316: We do not want to accept further messages if message_carried_over is not None

sentry_streams/src/batch_step.rs
```diff
@@ -0,0 +1,665 @@
...
+        &self,
+        committable: BTreeMap<Partition, u64>,
+    ) -> Message<RoutedValue> {
+        let ts = current_epoch();
+        let rv = RoutedValue {
+            route: self.route.clone(),
+            payload: RoutedValuePayload::make_watermark_payload(committable.clone(), ts),
+        };
+        Message::new_any_message(rv, committable)
+    }
```
fpacifici: Please inline this in emit_watermark_tail. It is used only there

sentry_streams/src/batch_step.rs
```diff
@@ -0,0 +1,715 @@
...
+        }
+        let inner = message.into_payload();
+        let pysm = match inner.payload {
+            RoutedValuePayload::PyStreamingMessage(m) => m,
+            _ => unreachable!(),
+        };
+        let PyStreamingMessage::PyAnyMessage { content } = pysm else {
+            unreachable!();
+        };
+        self.elements.push(content);
```
fpacifici: Let's make a change here and in the constructor. 
If we passed a PyAnyMessage and the Committable to append and to the init method, we would not need to manage these unreachable scenarios. The BatchStep methods would have already excldued the cases where we do not receive a PyAnyMessage. 

So
`append(&mut self, message: PyAnyMessage, committable: BTreeMap<Partition, u64>)`

You can adjust the types to ensure the borrow checker works.

sentry_streams/src/batch_step.rs
```diff
@@ -0,0 +1,715 @@
...
+        }
+        false
+    }
+
+    pub fn current_offsets_snapshot(&self) -> BTreeMap<Partition, u64> {
+        self.batch_offsets.clone()
+    }
+
+    /// GIL: build list payload from stored `Py<PyAnyMessage>` elements, wrap in a batched `Message`.
+    pub fn build_stacked_message(&self) -> Result<Message<RoutedValue>, StrategyError> {
```
fpacifici: Let's rename to `flush`

sentry_streams/src/batch_step.rs
```diff
@@ -0,0 +1,715 @@
...
+                merge_commit_request(self.commit_request_carried_over.take(), c);
+            match self.next_step.submit(msg) {
+                Ok(()) => {}
+                Err(SubmitError::MessageRejected(MessageRejected { message })) => {
+                    self.message_carried_over = Some(message);
+                }
+                Err(SubmitError::InvalidMessage(e)) => return Err(e.into()),
+            }
+        }
+        Ok(())
```
fpacifici: I think this whole logic can be simplified:
Now we manage these scenarios:
- we keep a carried over message to be pushed the next time separately from the list of watermark to push through
- we have a separate method to retry the downstream and from pushing watermark.

In theory watermarks, the synthetic watermark and the batch are all messages (message<RoutedValue>).
So we can keep a single list of the messages to send through.
It can work this way:

- Create a Vec of Message<RoutedValue> to push to the next step 
- When we have a batch to push or we want to flush watermarks we append them to the list
- At every poll we try to submit all the messages.
- If we receive a MessageRejected error we stop, put back the message on the list, wait for the next call to poll, return the highest commit of the successfully sent messages. 
- We do the same on join.

sentry_streams/src/batch_step.rs
```diff
@@ -0,0 +1,696 @@
...
+use std::time::{Duration, SystemTime, UNIX_EPOCH};
+
+fn invalid_message_submit_error(message: &Message<RoutedValue>) -> SubmitError<RoutedValue> {
+    match &message.inner_message {
+        InnerMessage::BrokerMessage(broker_message) => {
+            SubmitError::InvalidMessage(broker_message.into())
+        }
+        InnerMessage::AnyMessage(..) => {
+            panic!("BatchStep: invalid message on AnyMessage; Arroyo cannot surface InvalidMessage")
+        }
```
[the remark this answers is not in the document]
fpacifici: Broadcast is not supported now.

sentry_streams/src/batch_step.rs
```diff
@@ -0,0 +1,696 @@
...
+    fn drain_outbound(&mut self) -> Result<(), StrategyError> {
+        while let Some(msg) = self.outbound.pop_front() {
+            let c = self.next_step.poll()?;
+            self.commit_request_carried_over =
+                merge_commit_request(self.commit_request_carried_over.take(), c);
+            match self.next_step.submit(msg) {
+                Ok(()) => {
+                    if self.pending_batch {
+                        self.pending_batch = false;
+                    }
```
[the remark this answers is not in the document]
fpacifici: It is alright in this scenario, we jsut do not want to accumulate more than one batch

sentry_streams/src/batch_step.rs
```diff
@@ -0,0 +1,696 @@
...
+        }
+        false
+    }
+
+    pub fn current_offsets_snapshot(&self) -> BTreeMap<Partition, u64> {
+        self.batch_offsets.clone()
+    }
+
+    pub fn flush(&self) -> Result<Message<RoutedValue>, StrategyError> {
+        if self.elements.is_empty() {
```
evanh: Is this even possible? Could this no-op instead of returning an error?

sentry_streams/src/batch_step.rs
```diff
@@ -0,0 +1,696 @@
...
+        let wm_after_batch: Vec<_> = std::mem::take(&mut self.watermark_buffer);
+
+        self.outbound.push_back(batch_msg);
+        self.pending_batch = true;
+        self.enqueue_watermark_tail(wm_after_batch, committable_for_synthetic);
+        Ok(())
+    }
+
+    fn enqueue_watermark_tail(
+        &mut self,
```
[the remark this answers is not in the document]
evanh: This is an edge case but is fair.

evanh (approved):
Couple comments but overall seems good to me.

sentry_streams/src/batch_step.rs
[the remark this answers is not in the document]
fpacifici: IT cannot happen. `remaining` is already clamped to 0 
```
/// This will be [`Duration::ZERO`] if the deadline has elapsed
```

PR #309, opened 2026-05-01
feat(rust_streams): PipelineStats and Batch/Filter metrics

fpacifici:
## Summary

We are not collecting metrics from Rust Steps. This adds the equivalent abstractions we have in python.
The next step will be to replace the python ones with the rust ones.

Adds a Rust `pipeline_stats` module aligned with Python `PipelineStats` (`stats.py`): in-memory aggregation of per-step exec counts, error counts, and max duration per flush window (10s), then emission through the `metrics` crate only (`streams.pipeline.input.messages`, `errors`, `duration` with a `step` label). No Arroyo types on the recording path.

## Changes

- **RuntimeOperator**: `step_name` on `Batch` and `Filter`; Python adapter passes pipeline step names.
- **BatchStep**: `step_exec` per ingested streaming row; `step_timing` on batch flush (`b.flush()`).
- **Filter**: `step_exec` / `step_timing` / `step_error` around the Python predicate; PredicateFilter closure no longer calls Python `input_metrics` / `output_metrics` to avoid double counting.
- **Tests**: Rust unit tests for flush behavior and throttle (local metrics recorder).


sentry_streams/src/pipeline_stats.rs
```diff
@@ -0,0 +1,352 @@
...
+/// Aggregates per-step exec / error / timing and flushes periodically to DogStatsD via `metrics`.
+pub struct PipelineStats {
+    inner: Mutex<Inner>,
+    flush_interval: Duration,
+}
+
+impl PipelineStats {
+    fn with_flush_interval(flush_interval: Duration) -> Self {
+        Self {
+            inner: Mutex::new(Inner::new()),
```
markstory: Why do we need the additional buffering? Couldn't the metrics backend take care of this?

fpacifici: We were used to have that and it turned out to be a major performance issue (addressed here in the python version https://github.com/getsentry/streams/pull/305).

In a nutshell:
- multiple stats are collected at every step of the pipeline, including steps that are trivial, this makes such collection extremely high throughput.
- The metrics backend does additional work (see the profile attached to the linked PR): like tags normalization as it has to be general. It does not know there is one tag only, that ended up being the culprit of the overhead.
- Stats are much more lightweight. At every call we just bump a counter, which is viable.

Now, the rust version is in a mutex which may turn out to be a problem. I will profile that part, though the reason we need a light weight version of the buffered metrics is the one described in the python PR, we did not retrofit it in the buffer because we get speed in exchange for flexibility in the stats abstraction, which we want to preserve in the backend.

sentry_streams/src/header_filter_step.rs
```diff
@@ -85,22 +89,28 @@ impl ProcessingStrategy<RoutedValue> for HeaderIntEqualityFilter {
     }
 
     fn submit(&mut self, message: Message<RoutedValue>) -> Result<(), SubmitError<RoutedValue>> {
-        if self.route != message.payload().route {
+        if self.route != message.payload().route || message.payload().payload.is_watermark_msg() {
```
fpacifici, line 92: Moved here in order not to count watermarks in metrics

evanh: Should we add a tag or use a different metric for watermarks? I would think we'd still like metrics on watermark messages, but tracked separately from main messages.

evanh (approved):
Generally approved, added one comment.

sentry_streams/src/header_filter_step.rs
fpacifici: I think this is a good point. More importantly we should emit metrics when they are emitted and when they are committed. 

Let me think over it. 
I will send a PR for those that covers the overall pipeline 
https://linear.app/getsentry/issue/STREAM-921/add-metrics-to-track-watermarks-per-step

PR #310, opened 2026-05-04
feat(watermarks): add test for watermarks passing thru PythonAdapter

bmckerry:
Ticket: https://linear.app/getsentry/issue/STREAM-508/add-pythonadapter-to-rust-arroyo-watermark-unit-test

This PR adds a test for ensuring that pipeline watermarks and associated commit requests pass through a PythonAdapter step properly.

### DIFF
--- a/sentry_streams/src/watermark.rs
+++ b/sentry_streams/src/watermark.rs
@@ -245,4 +245,86 @@ mod tests {
             set_timestamp(0);
         });
     }
+
+    #[test]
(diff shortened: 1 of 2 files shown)


george-sentry (approved):
Test is similar to `test_watermark_to_commit_step` but uses `RuntimeOperator::PythonAdapter` instead of `RuntimeOperator::Map`. Since it passes, I approve.

PR #313, opened 2026-05-07
feat(k8s): Add optional canary Deployment to PipelineStep

fpacifici:
The `PipelineStep` sentry-kube macro can optionally emit a stable Deployment and a single-replica canary Deployment so pipeline steps can run split replicas while sharing one ConfigMap.

**`with_canary` and replica split**

When `with_canary` is true and `replicas` is at least two, the main manifest uses `replicas - 1` and the canary uses one replica. The canary resource name gets a `-canary` suffix and carries `env: canary` on deployment metadata labels, pod template labels, and `spec.selector.matchLabels` so callers can target it separately. If `replicas` is one with canary enabled, the macro emits only the main Deployment at full replica count and omits `canary_deployment`.

**Return shape and patches**

The macro still returns `deployment` and `configmap`. When canary splitting applies, it also returns `canary_deployment`, which Jinja or other renderers should emit as a second manifest. An `emergency_patch`, when provided, is merged into both Deployments.

### DIFF
--- a/sentry_streams_k8s/sentry_streams_k8s/pipeline_step.py
+++ b/sentry_streams_k8s/sentry_streams_k8s/pipeline_step.py
@@ -531,7 +531,7 @@ class PipelineStep(ExternalMacro):
                 emergency_patch=emergency_patch,
                 deployment_name=main_deployment_name,
                 replica_count=replicas,
-                step_labels=labels,
+                step_labels={**labels, "env": "primary"},
                 container=container,
                 volumes=volumes,
             )
--- a/sentry_streams_k8s/sentry_streams_k8s/pipeline_step.py
+++ b/sentry_streams_k8s/sentry_streams_k8s/pipeline_step.py
@@ -232,7 +232,56 @@ def parse_context(context: dict[str, Any]) -> PipelineStepContext:
         "emergency_patch": emergency_patch_parsed,
         "enable_liveness_probe": context.get("enable_liveness_probe", True),
         "container_name": context.get("container_name", "pipeline-consumer"),
+        "with_canary": bool(context.get("with_canary", False)),
     }
+
+
+def _build_merged_pipeline_deployment(
+    *,
+    base_deployment: dict[str, Any],
+    deployment_template: dict[str, Any],
+    emergency_patch: dict[str, Any],
+    deployment_name: str,
+    replica_count: int,
+    step_labels: dict[str, Any],
+    container: dict[str, Any],
+    volumes: list[dict[str, Any]],
+) -> dict[str, Any]:
+    pipeline_additions: dict[str, Any] = {
+        "metadata": {
+            "name": deployment_name,
+            "labels": step_labels,
+        },
+        "spec": {
+            "replicas": replica_count,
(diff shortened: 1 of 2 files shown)


fpacifici:
https://linear.app/getsentry/issue/STREAM-714/make-the-macro-generate-the-canary-deployment

sentry_streams_k8s/sentry_streams_k8s/pipeline_step.py
```diff
@@ -432,48 +493,46 @@ def run(self, context: dict[str, Any]) -> dict[str, Any]:
...
+            deployment = _build_merged_pipeline_deployment(
+                base_deployment=base_deployment,
+                deployment_template=deployment_template,
+                emergency_patch=emergency_patch,
+                deployment_name=main_deployment_name,
+                replica_count=replicas - 1,
+                step_labels=labels,
+                container=container,
+                volumes=volumes,
+            )
```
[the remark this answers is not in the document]
evanh: I'm not sure if this is a valid bug or not.

evanh (approved):

sentry_streams_k8s/sentry_streams_k8s/pipeline_step.py
```diff
@@ -232,9 +232,66 @@ def parse_context(context: dict[str, Any]) -> PipelineStepContext:
...
+    replica_count: int,
+    step_labels: dict[str, Any],
+    container: dict[str, Any],
+    volumes: list[dict[str, Any]],
+) -> dict[str, Any]:
+    """
+    Assembles a k8s deployment by layering these structures on top of the base deployment
+    manifest:
+    1. deployment_template: provided by the user
+    2. the steraming platform specific additions (including the container)
```
markstory:
```suggestion
    2. the streaming platform specific additions (including the container)
```

sentry_streams_k8s/sentry_streams_k8s/pipeline_step.py
```diff
@@ -432,48 +493,46 @@ def run(self, context: dict[str, Any]) -> dict[str, Any]:
...
-        if emergency_patch:
-            deployment = deepmerge(deployment, emergency_patch)
+        if add_canary:
+            deployment = _build_merged_pipeline_deployment(
+                base_deployment=base_deployment,
+                deployment_template=deployment_template,
+                emergency_patch=emergency_patch,
+                deployment_name=main_deployment_name,
+                replica_count=replicas - 1,
+                step_labels=labels,
```
markstory: Should these labels have `env: primary` added to separate them from the canary deployments?

fpacifici: Indeed it should. 
Turns out sbc does not do that and the main deployment also redeploys canary.
I geuss it is a mistake

sentry_streams_k8s/sentry_streams_k8s/pipeline_step.py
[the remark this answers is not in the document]
fpacifici: fixed

PR #314, opened 2026-05-07
feat(streams): Consumer watermark commit latency metric

fpacifici:
Adds a latency metrics to the consumer.
As we rely on watermarks for commit we can rely on watermarks for latency making the 
product oblivious to the generation of the latency metric.
Specifically:
- The watermark emitter adds the timestamp of the last message to the watermark
- The commit policy reads the message timestamp in every watermark and record the time between the message and now.
- Steps that generate synthetic watermark decide which timestamp to apply. Example: the batch step append the timestamp of the oldest message being flushed.


sentry_streams/src/commit_policy.rs
```diff
@@ -59,6 +64,14 @@ impl WatermarkCommitOffsets {
...
                 commit_request =
                     merge_commit_request(Some(commit_request), Some(current_request)).unwrap();
+
+                let now = current_epoch();
+                let secs = watermark
+                    .message_time
+                    .map(|t| now.saturating_sub(t) as f64)
+                    .unwrap_or(0.0);
+                metrics::histogram!(METRIC_WATERMARK_COMMIT_LATENCY).record(secs);
+
```
fpacifici, line 81: We want to record the latency only once per call to `commit`.
This logic should go below right before we return. 
AS many committables are merged together, we want to record the latency based on the oldest watermark, which is the one with the lowest message_time.

sentry_streams/src/commit_policy.rs
```diff
@@ -59,6 +64,14 @@ impl WatermarkCommitOffsets {
                 };
                 commit_request =
                     merge_commit_request(Some(commit_request), Some(current_request)).unwrap();
+
+                let now = current_epoch();
+                let secs = watermark
+                    .message_time
+                    .map(|t| now.saturating_sub(t) as f64)
+                    .unwrap_or(0.0);
+                metrics::histogram!(METRIC_WATERMARK_COMMIT_LATENCY).record(secs);
```
fpacifici: Please modify an existing test to validate the latency metrics is reported. 
See how the pipeline_stats.rs file manages to test metrics by creating a Recorder https://github.com/getsentry/streams/blob/main/sentry_streams/src/pipeline_stats.rs#L155

sentry_streams/sentry_streams/adapters/arroyo/rust_arroyo.py
```diff
@@ -81,7 +81,13 @@
 
 
 def build_py_metrics_config(cfg: MetricsConfig) -> PyMetricConfig | None:
-    """Build Rust-side DogStatsD config from the same metrics dict used by configure_metrics."""
+    """Build Rust-side DogStatsD config from the same metrics dict used by configure_metrics.
+
+    When this is installed, the consumer also emits
+    ``streams.pipeline.consumer.watermark_commit_latency`` (histogram, seconds): wall time from
+    optional watermark ``message_time`` (epoch seconds of the last or oldest batched data row) to
+    the commit decision; 0 is recorded when ``message_time`` is absent.
```
fpacifici: Please remove this

sentry_streams/sentry_streams/rust_streams.pyi
```diff
@@ -200,8 +200,11 @@ class PyWatermark:
         self,
         payload: dict[tuple[str, int], int],
         timestamp: int,
+        message_time: int | None = None,
```
fpacifici: We consistently use floats 64 for message timestamp. Please adjust message_time to that data type both in rust and python. 
see messages.rs as well.
No need to update the timestamp field in the Watermark and PyWatermark objects. That would be an unrelated change.

sentry_streams/src/messages.rs
```diff
@@ -388,11 +415,33 @@ impl RoutedValuePayload {
...
+
+/// For a routed broker/data message, the logical message time used for watermark `message_time`.
+pub fn data_message_time_secs_from_routed(message: &Message<RoutedValue>) -> Option<u64> {
+    let rv = message.payload();
+    match &rv.payload {
+        RoutedValuePayload::WatermarkMessage(_) => None,
+        RoutedValuePayload::PyStreamingMessage(ref sm) => Some(traced_with_gil!(|py| {
+            streaming_payload_timestamp_floor_secs(py, sm)
+        })),
     }
```
fpacifici: We should manage the time as float64. So these should not be needed.

sentry_streams/src/messages.rs
```diff
@@ -154,15 +173,23 @@ pub struct PyWatermark {
     pub committable: Py<PyDict>,
     #[pyo3(get)]
     pub timestamp: Py<PyInt>,
+    #[pyo3(get)]
+    pub message_time: Option<Py<PyInt>>,
```
fpacifici: This will have to be float

sentry_streams/src/commit_policy.rs
```diff
@@ -52,13 +60,25 @@ impl WatermarkCommitOffsets {
...
+        let mut oldest_message_time: Option<f64> = None;
         for (ts, watermark) in self.watermarks.iter() {
             if watermark.num_watermarks == self.num_branches {
                 let current_request = CommitRequest {
                     positions: watermark.committable.clone(),
                 };
                 commit_request =
                     merge_commit_request(Some(commit_request), Some(current_request)).unwrap();
+
+                if let Some(t) = watermark.message_time {
```
evanh: Nit: Calling this field `message_time` feels a little confusing. At first glance it seems like this is the time of the watermark itself, instead of the previous actual message.

evanh (approved):

sentry_streams/src/commit_policy.rs
fpacifici: replaced

sentry_streams/src/commit_policy.rs
```diff
@@ -71,6 +91,10 @@ impl WatermarkCommitOffsets {
         }
 
         if commit_request != empty_commit_request {
+            let secs = oldest_last_message_time
+                .map(|t| ((current_epoch() as f64) - t).max(0.0))
+                .unwrap_or(0.0);
+            metrics::histogram!(METRIC_WATERMARK_COMMIT_LATENCY).record(secs);
```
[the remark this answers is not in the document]
fpacifici: that's alright, we do not care too much about anything lower than seconds

PR #316, opened 2026-05-09
Add sentry SDK to the rust arroyo adapter

fpacifici:
We integrated the sentry sdk in the streaming pipeline code and the adapter
code but that does not cover anything happening in rust.
1. The rust consumer catches any panic and swallows them. So these errors do not go to the python code
2. Anyway when an arroyo strategy panics nothing is surfaced to send to python. See StrategyPanic
3. Also the rust_arroyo adapter does not crash if the consumer fails. It termiantes with error code 0.

This does not address step 3, but it introduces a sentry integration in the rust code.
The guard is created at the beginning of the run method.
The integration can be configured separately from the main pipeline sdk.
This separation gives us some ways to separate what goes wrong in the application from
what goes on in the adapter code.
It is by no means a precise separation but we can start trying this way.

In the current state:
- IF the product code rasies an error we would see it in both integrations
- If the strategy fails in the rust code, the error would go to the adapter integration only.


george-sentry (approved):

sentry_streams/src/consumer.rs
```diff
@@ -147,6 +153,16 @@ impl ArroyoConsumer {
 
         configure_metrics(self.metric_config.clone());
 
+        let _sentry_guard = self.sentry_dsn.as_ref().map(|dsn| {
+            sentry::init((
+                dsn.as_str(),
+                sentry::ClientOptions {
+                    release: sentry::release_name!(),
+                    ..Default::default()
```
markstory, line 161: Do you need to set the environment as well?

fpacifici: right. Will do

fpacifici: It is actually supposed to be already set as we provide the SENTRY_ENVIRONMENT envvar in all our prod systems.

PR #337, opened 2026-06-25
Bump kafka schemas

fpacifici:
Unfortunately we need to bump the library to update schemas as we have not done this
https://linear.app/getsentry/issue/STREAM-107/decouple-sentry-stream-from-sentry-kafka-schemas


bmckerry (approved):

untitaker (approved):
something weird is going on in the uv lockfile. why are so many transient deps bumped?

PR #339, opened 2026-07-01
feat(operator): initial operator implementation (no deploy)

bmcquilkin-sentry:
Phase 1 of creating a k8s operator to the streams repository to manage Kafka consumer pod lifecycles. Creates the StreamingConsumer CRD and a basic operator that reconciles resources into a Deployment and ConfigMap. Splits shared logic from that and the PipelineStep macro into consumer builder methods.


sentry_streams_k8s/sentry_streams_k8s/operator/manifests/operator/clusterrole.yaml
```diff
@@ -0,0 +1,17 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: streams-k8s-operator
+rules:
+  - apiGroups: ["streams.sentry.io"]
```
george-sentry: This seems specific to Sentry. Is that okay even though `streams` itself is open source?

sentry_streams_k8s/sentry_streams_k8s/operator/operator.py
```diff
@@ -0,0 +1,72 @@
...
+from sentry_streams_k8s.operator.streaming_consumer import (
+    from_crd_spec,
+    render,
+    validate,
+)
+
+GROUP = "streams.sentry.io"
+VERSION = "v1"
+PLURAL = "streamingconsumers"
+DEFAULTS_PATH_ENV = "STREAMS_K8S_OPERATOR_DEFAULTS"
```
george-sentry: If you plan on adding more constants in the future, it may be useful to put them in a separate file. Not very important right now though.

sentry_streams_k8s/sentry_streams_k8s/operator/operator.py
```diff
@@ -0,0 +1,72 @@
...
+    YAML file named by STREAMS_K8S_OPERATOR_DEFAULTS (empty when unset)."""
+    global _operator_defaults
+    if _operator_defaults is None:
+        path = os.environ.get(DEFAULTS_PATH_ENV)
+        if path:
+            with open(path) as f:
+                _operator_defaults = yaml.safe_load(f) or {}
+        else:
+            _operator_defaults = {}
+    return _operator_defaults
```
george-sentry: It's interesting that `_operator_defaults` is global, but you also return it from `load_operator_defaults`. Why not simply build it inside `load_operator_defaults` and avoid the global variable?

sentry_streams_k8s/sentry_streams_k8s/operator/operator.py
```diff
@@ -0,0 +1,72 @@
...
+
+
+@kopf.on.create(GROUP, VERSION, PLURAL)
+@kopf.on.update(GROUP, VERSION, PLURAL)
+@kopf.on.resume(GROUP, VERSION, PLURAL)
+def reconcile(spec: kopf.Spec, name: str, namespace: str | None, **_: Any) -> None:
+    assert namespace is not None
```
george-sentry, line 75: This `assert` suggests to me that `namespace` can never be None, and if it is, there is something seriously broken with the system.

If it's possible for `namespace` to be None under normal circumstances, maybe it would be better to raise an error that says it must be provided and then exit?

sentry_streams_k8s/sentry_streams_k8s/operator/streaming_consumer.py
```diff
@@ -0,0 +1,287 @@
...
+    auto_offset_reset: NotRequired[str]
+    override_params: NotRequired[dict[str, Any]]
+
+
+class ResourcesSpec(TypedDict):
+    cpu_per_process: int
+    memory_per_process: int
+
+
+class OperatorDefaults(TypedDict, total=False):
```
george-sentry: Is this class actually used anywhere? It is one of the possible types for the `default` argument in `from_crd_spec`, but it is only ever called with the output of `load_operator_defaults`, which is just a plain dictionary.

sentry_streams_k8s/sentry_streams_k8s/operator/streaming_pipeline.py
```diff
@@ -0,0 +1,287 @@
...
+REQUIRED_KAFKA_FIELDS = ("consumer_group", "bootstrap_servers")
+
+
+def validate(spec: StreamingConsumerSpec) -> None:
+    """Check required merged fields, then defer to ``PipelineStep.validate_context``
+    for the same schema/liveness checks the macro applies."""
+    missing = [f for f in REQUIRED_FIELDS if f not in spec]
+    if "steps" not in spec:
+        missing.append("steps")
+    if missing:
```
george-sentry, line 49: If `StreamingConsumerSpec` is missing required fields, then it's not a `StreamingConsumerSpec`.

Here's what might be a bit better - `validate` takes a plain object or dictionary, and returns either a valid `StreamingConsumerSpec` or raises an error.

One of the risks of this current form of `validate` is that you may call it on a possibly valid or invalid spec, and then forget later whether your `StreamingConsumerSpec` was validated or not.

fpacifici:
Please provide a description for the PR with context.

fpacifici:
Also please update this file https://github.com/getsentry/streams/tree/main/sentry_streams_k8s#readme
If you add the operator there, the readme is not accurate anymore.

sentry_streams_k8s/sentry_streams_k8s/operator/manifests/operator/clusterrole.yaml
fpacifici, line 1: I do not think the manifest to deploy the operator belongs in the library.
This includes:
clusterrole.yaml
clusterrolebinding.yaml
deployment.yaml
serviceaccount.yaml
namespace.yaml.
This excludes the CRD.

This is because everything related with the operator deployment is specific to the clsuter we deploy that in. Examples:
- we do not want to re-release the library to be able to change the resources, or the namespace
- The generation of the namespace belongs to the ops repo
- The place we deploy the operator can expect security parameters or network policies.

So, I think you should do this:
1. At this stage remove the files above from here. You can keep the CRD
2. When we deploy them we add the files to the ops repo in your service.
3. Later we can embed the operator deployment in a custom macro (like the one we have now for streams)

sentry_streams_k8s/sentry_streams_k8s/operator/manifests/crd.yaml
```diff
@@ -0,0 +1,125 @@
...
+      additionalPrinterColumns:
+        - name: Pipeline
+          type: string
+          jsonPath: .spec.pipeline_name
+        - name: Replicas
+          type: integer
+          jsonPath: .spec.replicas
+        - name: Age
+          type: date
+          jsonPath: .metadata.creationTimestamp
```
fpacifici, line 27: WHat is this for ?

sentry_streams_k8s/sentry_streams_k8s/operator/manifests/crd.yaml
```diff
@@ -0,0 +1,125 @@
...
+                kafka:
+                  type: object
+                  required: [consumer_group, bootstrap_servers]
+                  properties:
+                    consumer_group:
+                      type: string
+                    bootstrap_servers:
+                      type: array
+                      items:
+                        type: string
```
fpacifici: I am not sure I fully understand this part. 
This is an example of what we pass to the macro 
https://github.com/getsentry/ops/blob/master/k8s/services/super-big-consumers/_items_metric_stream.yaml#L20-L34

As you may see, between other parameters, we pass the pipeline config from its config file. That is an input to your system (the rendered config), not something you have to build.

Example:
https://github.com/getsentry/ops/blob/master/k8s/services/super-big-consumers/_items_log_config.yaml
This is the config file. Your operator does not need to build this, it will be provided as input.

sentry_streams_k8s/sentry_streams_k8s/operator/manifests/crd.yaml
```diff
@@ -0,0 +1,125 @@
...
+                  default: true
+                with_canary:
+                  type: boolean
+                  default: false
+                service_account:
+                  type: string
+                node_selector:
+                  type: object
+                  additionalProperties:
+                    type: string
```
fpacifici: THis should not be input.
Please pass a deployment template to customzie the deployment, just as we do for the macro:
https://github.com/getsentry/ops/blob/master/k8s/services/super-big-consumers/_items_log_stream.yaml#L12-L18

Deployment template and container template are properties of the environment you run the operator in. That is the way the user will customize their deployment. The system will be a lot more flexible if we allowed the client to specify a template rather than allowing specific fields to be provided.

sentry_streams_k8s/sentry_streams_k8s/operator/operator.py
```diff
@@ -0,0 +1,72 @@
...
+def load_operator_defaults() -> dict[str, Any]:
+    """Service/region defaults merged under every CR .spec, read once from the
+    YAML file named by STREAMS_K8S_OPERATOR_DEFAULTS (empty when unset)."""
+    global _operator_defaults
+    if _operator_defaults is None:
+        path = os.environ.get(DEFAULTS_PATH_ENV)
+        if path:
+            with open(path) as f:
+                _operator_defaults = yaml.safe_load(f) or {}
+        else:
```
fpacifici: What are the defaults we are passing as env vars and why env vars ?

sentry_streams_k8s/sentry_streams_k8s/operator/streaming_consumer.py
```diff
@@ -0,0 +1,287 @@
...
+    node_selector: NotRequired[dict[str, str]]
+
+    enable_kafka_auth: NotRequired[bool]
+    kafka_sasl_username: NotRequired[str]
+    kafka_sasl_secret_name: NotRequired[str]
+
+    emergency_patch: NotRequired[dict[str, Any]]
+
+
+DEFAULT_SERVICE_ACCOUNT = "service-super-big-consumer"
```
fpacifici: Please do not have a default for this. The user has to specify that. super-big-consumer is something specific to our system.

sentry_streams_k8s/sentry_streams_k8s/operator/streaming_consumer.py
```diff
@@ -0,0 +1,287 @@
...
+
+
+def _kafka_override_params(spec: StreamingConsumerSpec) -> dict[str, Any]:
```
fpacifici: AS tkafka parameters are part of the config yaml you should not paply overrides here.

sentry_streams_k8s/sentry_streams_k8s/operator/streaming_consumer.py
```diff
@@ -0,0 +1,287 @@
...
+    username = spec.get("kafka_sasl_username", DEFAULT_KAFKA_SASL_USERNAME)
+    return {
+        "security.protocol": DEFAULT_KAFKA_SECURITY_PROTOCOL,
+        "sasl.mechanism": DEFAULT_KAFKA_SASL_MECHANISM,
+        "sasl.username": username,
+        "sasl.password": "${envvar:KAFKA_SASL_PASSWORD}",
+    }
+
+
+def build_pipeline_config(spec: StreamingConsumerSpec) -> dict[str, Any]:
```
fpacifici: As explained above, config is not something you render, it is an input to your operator. The operator will make it a configmap

sentry_streams_k8s/sentry_streams_k8s/operator/streaming_consumer.py
```diff
@@ -0,0 +1,287 @@
...
+        "tags": {"pipeline": metrics_tag},
+    }
+    config["pipeline"] = {
+        "adapter_config": {"arroyo": arroyo},
+        "segments": [{"steps_config": steps_config}],
+    }
+    return config
+
+
+def build_deployment_template(spec: StreamingConsumerSpec) -> dict[str, Any]:
```
fpacifici: Same as above, this is an input

sentry_streams_k8s/sentry_streams_k8s/operator/streaming_consumer.py
```diff
@@ -0,0 +1,287 @@
...
+            {
+                "name": "KAFKA_SASL_PASSWORD",
+                "valueFrom": {"secretKeyRef": {"name": secret_name, "key": "KAFKA_SASL_PASSWORD"}},
+            }
+        )
+    return {"env": env, "securityContext": dict(SECURITY_CONTEXT)}
+
+
+def _default_component(spec: StreamingConsumerSpec) -> str:
+    return f"sbc-streams-{spec['pipeline_name']}-gcs"
```
fpacifici: Please do not provide a default component. The user has to provide a component

sentry_streams_k8s/sentry_streams_k8s/operator/streaming_consumer.py
```diff
@@ -0,0 +1,287 @@
...
+
+
+def _default_app_feature(spec: StreamingConsumerSpec) -> str:
+    return f"analytics_{spec['pipeline_name'].replace('-', '_')}"
```
fpacifici: Same, there is no default for this.

sentry_streams_k8s/sentry_streams_k8s/operator/streaming_consumer.py
```diff
@@ -0,0 +1,287 @@
...
+        raise ValueError(f"StreamingConsumer .spec.kafka is missing: {', '.join(missing_kafka)}")
+
+    PipelineStep.validate_context(build_pipeline_step_context(spec))
+
+
+def render(spec: StreamingConsumerSpec) -> dict[str, Any]:
+    """Render to ``configmap`` + ``deployment`` (+ ``canary_deployment`` when
+    canary splitting is active), via ``PipelineStep.run``."""
+    context = build_pipeline_step_context(spec)
+    return PipelineStep().run(context)
```
fpacifici: I don't think dependencies here are in the right direciton.
The PR makes the operator depend on the macro which contains the application logic. This couples the operator to the macro a lot. 
If you have to make a change to the macro interface you break the operator and have to make changes there.

I's see the operator and the macro as two separate adapters on top of the business logic that generates the consumer. This allows you to put the consumer generation logic in its own module and avoid coupling the macro related logic, the oeprator related logic with our own business logic.

I would go this way

```mermaid
graph TD;
  Operator
  Macro
  ConsumerBuilder

  Operator -->   ConsumerBuilder
  Macro -->   ConsumerBuilder
```

fpacifici (changes requested):
Several high level changes:
- Let's break down this PR. I would recommend Putting together the CRD and an empty operator here. So the key of this PR is the CRD and we can iterate on that.
- Break down the macro to extract the business logic so you can use it from the operator as well. See my comments on the dependencies 
- Then we can start fill in the business logic in the operator

sentry_streams_k8s/sentry_streams_k8s/operator/manifests/crd.yaml
bmcquilkin-sentry: What outputs when you use `kubectl` to list the consumers

george-sentry (approved):

sentry_streams_k8s/pyproject.toml
```diff
@@ -16,21 +16,33 @@ classifiers = [
...
-    "requests>=2.32.3",
-    "sentry-infra-tools>=1.24.0",
     "sentry-streams>=0.0.40",
     "pyyaml>=6.0.2",
 ]
 
+[project.optional-dependencies]
+macro = [
+    "sentry-infra-tools>=1.24.0",
+]
```
fpacifici, line 26: Are you sure the pypi library is going to build successfully this way ?
This is what the build workflow does https://github.com/getsentry/streams/blob/main/.github/workflows/build-streams-k8s.yaml#L22.

sentry_streams_k8s/sentry_streams_k8s/operator/operator.py
```diff
@@ -0,0 +1,117 @@
...
+        kopf.adopt(manifest)
+        _apply(dyn, manifest, namespace)
+
+    _prune_stale_deployments(
+        namespace=namespace,
+        owner_uid=uid,
+        service_name=consumer["service_name"],
+        pipeline_name=consumer["pipeline_name"],
+        desired_names={m["metadata"]["name"] for m in manifests if m["kind"] == "Deployment"},
+    )
```
fpacifici, line 100: I don't think I understand this. 
Could you give me an example about when we would remove a deployment here ?

fpacifici (commented):
One last clarification.
What is the scenario where we would prune deployments ?

sentry_streams_k8s/sentry_streams_k8s/operator/operator.py
bmcquilkin-sentry: Changing `with_canary` from true to false would need this.

bmcquilkin-sentry: Also not sure if we would ever do this but in the case where we delete a CR we would need it.

fpacifici (approved):
I see the unit test are validating rendering. 
Do we have a way to test the behavior of the operator ? I think you should at least unit test the "reconcile" function against a mock of the Kubernetes api before merging.

PR #342, opened 2026-07-07
ref(k8s): build operator and push to gar

bmcquilkin-sentry:
Build the operator image and push to the Sentry GAR. Modeled after `sentry-kafka-management`.


untitaker (approved):

.github/workflows/build-operator.yaml
```diff
@@ -0,0 +1,40 @@
...
+      contents: read
+      packages: write
+      id-token: write
+    strategy:
+      matrix:
+        include:
+          - registry: single-region
+            image: us-central1-docker.pkg.dev/sentryio/streaming-operator/image
+          - registry: multi-region
+            image: us-docker.pkg.dev/sentryio/streaming-operator-mr/image
```
markstory, line 25: I don't think we need both. Only publishing to multi-region is good enough. Some applications do both as we had to move from single -> multi region.

sentry_streams_k8s/Dockerfile
```diff
@@ -0,0 +1,43 @@
+# syntax=docker/dockerfile:1
+
+FROM python:3.11-slim
+
+ENV PYTHONDONTWRITEBYTECODE=1 \
```
fpacifici: Why this ? It will be a problem if you start spawning processes as it would recompile everything

.github/workflows/build-operator.yaml
```diff
@@ -0,0 +1,40 @@
+name: Build & Push Streaming Operator Image
+
+on:
+  push:
+    branches:
+      - main
+      - release-sentry-streams-k8s/**
+    paths:
+      - sentry_streams_k8s/**
+      - .github/workflows/build-operator.yaml
```
fpacifici, line 9: What is this for ?

.github/workflows/build-operator.yaml
```diff
@@ -0,0 +1,40 @@
+name: Build & Push Streaming Operator Image
+
+on:
+  push:
+    branches:
+      - main
+      - release-sentry-streams-k8s/**
```
fpacifici: You do not need this. That is to release libraries. WE are just pushing the docker image here

fpacifici (approved):
Please address the comment in line before merging

PR #343, opened 2026-07-09
feat(k8s): package the operator as a helm chart

bmcquilkin-sentry:
Packages the operator as a Helm chart which includes the operator image and CRD for consumers.


sentry_streams_k8s/chart/streaming-operator/Chart.yaml
```diff
@@ -0,0 +1,10 @@
+# version and appVersion are kept equal to the sentry-streams-k8s python
+# package version by scripts/bump-version.sh. do not manually bump them.
```
fpacifici: Why do you want to keep them in sync? They are two completely different release processes. If we decided to keep them in sync we would need to release the docker image only when we ship the streams-k8s library that is unnecessary.

sentry_streams_k8s/chart/streaming-operator/Chart.yaml
fpacifici, line 1: Please add kubeVersion: ">=1.23.0-0"

sentry_streams_k8s/chart/streaming-operator/crds/crd.yaml
fpacifici, line 1: I think this is in the wrong place. 
CRDs should not be in templates but in `/crds` 
https://helm.sh/docs/topics/charts#the-chart-file-structure

sentry_streams_k8s/chart/streaming-operator/templates/deployment.yaml
```diff
@@ -0,0 +1,65 @@
+{{- if gt (int .Values.replicaCount) 1 }}
+{{- fail "streaming-operator supports a single replica only: kopf runs standalone without leader election, two replicas would double-reconcile" }}
+{{- end }}
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: {{ include "streaming-operator.fullname" . }}
+  namespace: {{ .Release.Namespace }}
+  labels:
```
fpacifici, line 6: We will need to add cogs labels here:
- app-feature: infra 
- system: k8s_operator

sentry_streams_k8s/chart/streaming-operator/templates/deployment.yaml
```diff
@@ -0,0 +1,65 @@
...
+    {{- include "streaming-operator.labels" . | nindent 4 }}
+spec:
+  replicas: {{ .Values.replicaCount }}
```
fpacifici: Why is this not hardocded to 1 now ?

sentry_streams_k8s/chart/streaming-operator/templates/rbac.yaml
```diff
@@ -0,0 +1,43 @@
...
+rules:
+  - apiGroups: [apiextensions.k8s.io]
+    resources: [customresourcedefinitions]
+    verbs: [get, list, watch]
+  - apiGroups: [""]
+    resources: [namespaces]
+    verbs: [get, list, watch]
+  - apiGroups: [streams.sentry.io]
+    resources: [streamingconsumers]
+    verbs: [get, list, watch, patch, update]
```
fpacifici: Why does this need to patch and update the resources of your CRD ?

sentry_streams_k8s/chart/streaming-operator/templates/rbac.yaml
```diff
@@ -0,0 +1,43 @@
...
+  - apiGroups: [streams.sentry.io]
+    resources: [streamingconsumers/status]
+    verbs: [get, patch, update]
```
fpacifici: What do you need for ?

sentry_streams_k8s/chart/streaming-operator/templates/rbac.yaml
```diff
@@ -0,0 +1,43 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: {{ include "streaming-operator.fullname" . }}
+  labels:
+    {{- include "streaming-operator.labels" . | nindent 4 }}
+rules:
+  - apiGroups: [apiextensions.k8s.io]
```
fpacifici, line 8: Please scope all of these to the `default` namespace. I know it does not remove much but at least it is a start.

sentry_streams_k8s/chart/streaming-operator/templates/rbac.yaml
fpacifici, line 1: Let's make it a bit more restricted now that we are building it for the first time. 
1. have a steraming-pipelines namespace
2. add the deployment, pod and configmap update operations only on that namespace
3. Have the read permissions on everything.

Basically we will create the instances of the CRD anywhere (default namespace probably) but the resources controlled by the operator will be all in one isolated namespace for now.

.github/workflows/publish-operator.yaml
```diff
@@ -0,0 +1,93 @@
+name: Publish Streaming Operator Release (image + helm chart)
+
+on:
+  push:
+    tags:
+      - "*"
```
fpacifici: Shouldn't this be only on main ?

.github/workflows/publish-operator.yaml
```diff
@@ -0,0 +1,93 @@
...
+  workflow_dispatch:
+
+concurrency:
+  group: publish-operator-${{ github.ref }}
+  cancel-in-progress: false
```
fpacifici: Why ?

.github/workflows/publish-operator.yaml
```diff
@@ -0,0 +1,93 @@
...
+            echo "publish=false" >> "$GITHUB_OUTPUT"
+            exit 0
+          fi
+          CHART_VERSION=$(awk '/^version:/ {print $2}' "${CHART_DIR}/Chart.yaml")
+          if [[ "${CHART_VERSION}" != "${VERSION}" ]]; then
+            echo "::error::Chart.yaml version (${CHART_VERSION}) != package version (${VERSION}); scripts/bump-version.sh should keep them equal."
+            exit 1
+          fi
+          echo "publish=true" >> "$GITHUB_OUTPUT"
+          echo "version=${VERSION}" >> "$GITHUB_OUTPUT"
```
fpacifici: I don't think we need to tie the chart release to the library one.
We release manually the library just because our toolchain requires it. 
WE do not need to do it for anything that is packaged as a Docker image

.github/workflows/publish-operator.yaml
```diff
@@ -0,0 +1,93 @@
...
+
+      - name: GCP auth
+        if: steps.vars.outputs.publish == 'true'
+        uses: google-github-actions/auth@7c6bc770dae815cd3e89ee6cdf493a5fab2cc093 # v3.0.0
+        with:
+          workload_identity_provider: "projects/868781662168/locations/global/workloadIdentityPools/prod-github/providers/github-oidc-pool"
+          service_account: "gha-gcr-push@sac-prod-sa.iam.gserviceaccount.com"
+
+      - name: Set up gcloud
```
fpacifici: Have you discussed this with prod eng about how to distribute the chart ?

fpacifici (changes requested):
Please see my comments  in line

sentry_streams_k8s/chart/streaming-operator/templates/rbac.yaml
gi0baro: -1. Should be scoped within the `Release.namespace`

sentry_streams_k8s/chart/streaming-operator/crds/crd.yaml
gi0baro: +1

sentry_streams_k8s/chart/streaming-operator/templates/deployment.yaml
```diff
@@ -0,0 +1,65 @@
+{{- if gt (int .Values.replicaCount) 1 }}
+{{- fail "streaming-operator supports a single replica only: kopf runs standalone without leader election, two replicas would double-reconcile" }}
+{{- end }}
```
gi0baro: If that's the case, why not hardcode the value in the chart and strip `replicaCount` from values?

sentry_streams_k8s/chart/streaming-operator/templates/deployment.yaml
gi0baro: @fpacifici no, we shouldn't hard-code labels in the templates, they should always come from values

sentry_streams_k8s/chart/streaming-operator/templates/deployment.yaml
gi0baro: +1

sentry_streams_k8s/chart/streaming-operator/Chart.yaml
gi0baro: +1

gi0baro (commented):
A few things:
- on the release process, there's already a chart release action (https://github.com/helm/chart-releaser-action), I would use that instead of re-inventing the wheel/process (example usage: https://github.com/gi0baro/tfstater/blob/master/.github/workflows/helm.yml)
- I'd suggest to keep helpers as generic as possible (you can take a look at our internal relay chart for inspiration (https://github.com/getsentry/ops/tree/master/k8s/helm_services/relay/chart)
- I'd go *fully-dynamic* in the deployment: everything comes from values but securityContext, I'd treat it the same way of everything else

Also, see my individual comments under @fpacifici ones

sentry_streams_k8s/chart/streaming-operator/templates/rbac.yaml
fpacifici: @gi0baro this is not going to be deployed in the same namespace as the consumers it manages. The operator is likely going to run into sentry-system, right ? Or is the recommendation different ?
The consumers would be in their own namespaces.
Regardless we should restrict the namespace whatever the one is. Probably from the value file

sentry_streams_k8s/chart/streaming-operator/templates/rbac.yaml
bmcquilkin-sentry: update is not needed but patch is needed by kopf

sentry_streams_k8s/chart/streaming-operator/crds/crd.yaml
bmcquilkin-sentry: I believe this does mean that CRD updates will require its own pipeline action since helm won't reapply it unless its in templates?

sentry_streams_k8s/chart/streaming-operator/templates/rbac.yaml
bmcquilkin-sentry: Noting that this requires us to manually implement DELETE since kopf.adopt + ownerReferences do not work when the CR and its objects are in different namespaces.

.github/workflows/publish-operator.yaml
bmcquilkin-sentry: If something is merged into main while another merge is still publishing (e.g. back to back merged) they both will not publish at the same time

sentry_streams_k8s/chart/streaming-operator/templates/rbac.yaml
fpacifici: > kopf.adopt + ownerReferences do not work when the CR and its objects are in different namespaces.

Could you expand on this ?

sentry_streams_k8s/chart/streaming-operator/templates/namespace.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: streaming-pipelines
```
fpacifici: The namespace has to be a value, it cannot be hardcoded.

sentry_streams_k8s/sentry_streams_k8s/operator/operator.py
```diff
@@ -19,37 +20,103 @@
 VERSION = "v1alpha1"
 PLURAL = "streamingpipelines"
 FIELD_MANAGER = "streaming-operator"
+WORKLOAD_NAMESPACE = "streaming-pipelines"
```
fpacifici: See above. This cannot be hardcoded

sentry_streams_k8s/sentry_streams_k8s/operator/operator.py
```diff
@@ -19,37 +20,103 @@
...
+            )
+
     dyn.server_side_apply(
         resource,
         body=manifest,
-        namespace=namespace,
-        field_manager=FIELD_MANAGER,
-        force_conflicts=True,
+        namespace=WORKLOAD_NAMESPACE,
+        field_manager=f"{FIELD_MANAGER}-{owner_uid}",
```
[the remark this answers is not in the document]
fpacifici: I think Cursor is right, though why do we need server side apply ?

sentry_streams_k8s/chart/streaming-operator/templates/rbac.yaml
bmcquilkin-sentry: Usually we do not need to implement the DELETE since we can use kopf.adopt + ownerReferences to link a Deployment to its CR. Then when the CR is deleted, any owned objects (like the Deployments) will be removed by k8s automatically. This only works when the CR and Deployment are in the same namespace, which is incorrect here since you are saying we could have a CR in `default` that creates a Deployment in `streaming-pipelines`.

fpacifici: CR would not be in default. IT would likely be in sentry-system

sentry_streams_k8s/sentry_streams_k8s/operator/operator.py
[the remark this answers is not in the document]
fpacifici: Please expand on why we need server_side_apply.

sentry_streams_k8s/chart/streaming-operator/templates/rbac.yaml
fpacifici: So what would the path be here ? I guess you would have to implement the deletion in the operator ?

sentry_streams_k8s/sentry_streams_k8s/operator/operator.py
```diff
@@ -107,6 +185,16 @@ def reconcile(
...
+    patch.status["workload_namespace"] = WORKLOAD_NAMESPACE
+
+
+@kopf.on.delete(GROUP, VERSION, PLURAL)
+def cleanup(uid: str, **_: Any) -> None:
+    _prune_stale_resources(
+        owner_uid=uid,
+        desired_deployments=set(),
+        desired_configmaps=set(),
+    )
```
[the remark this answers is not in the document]
fpacifici: I think CUrsor is right

fpacifici (approved):
Please address the comments and explain the server side apply before merging.

sentry_streams_k8s/sentry_streams_k8s/operator/operator.py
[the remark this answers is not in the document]
bmcquilkin-sentry: server_side_apply is used to update/create a resource (if it doesn't exist it will be created), and it will only update the fields that the CRD specifies.

sentry_streams_k8s/chart/streaming-operator/templates/rbac.yaml
bmcquilkin-sentry: Regardless of where it is we cannot use owner references, so yes we would have to just implement DELETE ourselves which is pretty trivial.

sentry_streams_k8s/sentry_streams_k8s/operator/operator.py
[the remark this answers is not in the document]
bmcquilkin-sentry: Not an issue, it will retry if the deletion fails.

PR #345, opened 2026-07-20
feat(operator): pod management

bmcquilkin-sentry:
Adds Pod management abilities to the operator. Instead of generating a Deployment, the operator will generate and manage Pods. Uses a ConfigMap per CR to track the Pod generations. Watches Pod events and uses a health classifier to determine how to respond to them and reconcile.

This is a different implementation than the original test we did and is based on Kopf's official suggestion for how to reconcile: https://docs.kopf.dev/en/latest/reconciliation/#level-based-triggering

For each StreamingPipeline CR, the operator starts a daemon in the same container that owns the reconciliation for that CR. Updates and events only set a flag that requests reconcile. The daemon runs a reconcile on one of these triggers or the 60s health check timer. Since we only set a flag, multiple events on the same CR will only trigger one reconcile (until it is processed). We also avoid the risk of multiple reconciles on the same CR.


sentry_streams_k8s/sentry_streams_k8s/operator/reconcile.py
```diff
@@ -0,0 +1,463 @@
...
+        ),
+    )
+    for configmap in configmaps.items:
+        if configmap.metadata.name not in desired_configmaps:
+            logger.info(
+                "Pruning stale configmap %s/%s", workload_namespace, configmap.metadata.name
+            )
+            core.delete_namespaced_config_map(
+                name=configmap.metadata.name, namespace=workload_namespace
+            )
```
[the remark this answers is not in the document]
bmcquilkin-sentry: This is real if we test the operator with Deployments on s4s2 before merging

sentry_streams_k8s/sentry_streams_k8s/operator/generations.py
```diff
@@ -0,0 +1,126 @@
+"""
+Pods for a replica are named {name}-{ordinal}-{generation}. The generation
+increments every time the operator replaces a Pod, so a newly created Pod
+never shares a name with the old one that is still terminating. This is what
+lets us recreate instantly instead of waiting for the old Pod to be deleted.
+We save the highest generation per ordinal in a ConfigMap (one per CR).
```
fpacifici: I am not sure you need a configMap to store the most recent generation.
It seems this could fit in the status of your CRD.
See details here https://kubernetes.io/docs/concepts/overview/working-with-objects/#object-spec-and-status

The `spec` field is what the manifest provides: what you want to achieve, the `status` field is managed by the operator/controller and represent data bout the current state of the resource. 
Example:
this is the status field of a statefulset
```
status:
  availableReplicas: 39
  collisionCount: 0
  currentReplicas: 39
  currentRevision: task-ingest-errors-broker-947979787
  observedGeneration: 150
  readyReplicas: 39
  replicas: 39
  updateRevision: task-ingest-errors-broker-947979787
  updatedReplicas: 39
```
You may see it contains a current revision.

I am pretty sure kopf allows you to set the current status of an object during the callbacks:
https://docs.kopf.dev/en/latest/walkthrough/updates/#updating-the-objects

So you should not have to manage this state on your own via a configmap

sentry_streams_k8s/sentry_streams_k8s/operator/pod_resources.py
```diff
@@ -0,0 +1,200 @@
...
+    pod_labels,
+    pod_metadata,
+)
+
+
+def _pod_resource(dyn: DynamicClient) -> Any:
+    return dyn.resources.get(api_version="v1", kind="Pod")
+
+
+def _to_dict(obj: Any) -> dict[str, Any]:
```
fpacifici: Why do you need to process the resources as dictionaries ?

sentry_streams_k8s/sentry_streams_k8s/operator/pod_resources.py
```diff
@@ -0,0 +1,200 @@
...
+    to_dict = getattr(obj, "to_dict", None)
+    if callable(to_dict):
+        return cast(dict[str, Any], to_dict())
+    return cast(dict[str, Any], client.ApiClient().sanitize_for_serialization(obj))
+
+
+def list_owned_pods(
+    dyn: DynamicClient,
```
fpacifici: Why do you need to use a Dynamic client to manage standard resources like pods ?

sentry_streams_k8s/sentry_streams_k8s/operator/pod_resources.py
```diff
@@ -0,0 +1,200 @@
...
+    namespace: str,
+    owner_uid: str,
+    workload_set: str | None = None,
```
fpacifici, line 32: What is a workload_set ?

sentry_streams_k8s/sentry_streams_k8s/operator/pod_resources.py
```diff
@@ -0,0 +1,200 @@
...
+    if workload_set is not None:
+        selector = f"{selector},{WORKLOAD_SET_LABEL}={workload_set}"
+    listing = _pod_resource(dyn).get(namespace=namespace, label_selector=selector)
+    return [_to_dict(item) for item in listing.items]
+
+
+def apply_pod(dyn: DynamicClient, pod: Mapping[str, Any], namespace: str) -> None:
+    _pod_resource(dyn).server_side_apply(
+        body=pod,
+        name=pod_metadata(pod)["name"],
```
fpacifici: You have a pod_name method but you are not using here. Any reason?

sentry_streams_k8s/sentry_streams_k8s/operator/pod_resources.py
```diff
@@ -0,0 +1,200 @@
...
+
+def pod_name(pod: Mapping[str, Any]) -> str:
+    return cast(str, pod_metadata(pod).get("name", ""))
+
+
+def consumer_pod_name(base_name: str, ordinal: int, generation: int) -> str:
+    return f"{base_name}-{ordinal}-{generation}"
+
+
+def pod_ordinal(pod: Mapping[str, Any]) -> int | None:
```
fpacifici: What's a pod ordinal ?

sentry_streams_k8s/sentry_streams_k8s/operator/pod_resources.py
```diff
@@ -0,0 +1,200 @@
...
+    label = pod_labels(pod).get(GENERATION_LABEL)
+    if label is None:
+        return 0
+    try:
+        return int(label)
+    except ValueError:
+        return 0
+
+
+def pod_workload_set(pod: Mapping[str, Any]) -> str | None:
```
fpacifici: What's a workload set ?

sentry_streams_k8s/sentry_streams_k8s/operator/pod_resources.py
```diff
@@ -0,0 +1,200 @@
...
+        "kind": "Pod",
+        "metadata": {
+            "name": consumer_pod_name(base_name, ordinal, generation),
+            "labels": labels,
+            "annotations": annotations,
+        },
+        "spec": pod_spec,
+    }
+
+    pod["metadata"]["annotations"][SPEC_HASH_ANNOTATION] = _pod_spec_hash(pod)
```
fpacifici: What do you need this hash annotation for ?

sentry_streams_k8s/sentry_streams_k8s/operator/pod_resources.py
```diff
@@ -0,0 +1,200 @@
...
+    OWNER_UID_LABEL,
+    SPEC_HASH_ANNOTATION,
+    WORKLOAD_SET_LABEL,
+    Logger,
+)
+from sentry_streams_k8s.operator.pod_health import (
+    is_deleting,
+    pod_labels,
+    pod_metadata,
+)
```
fpacifici: I am having a hard time following the design and the dependencies here.
It seems to me this module would provide an interface to manage the manifest of a pod and it should be imported by other modules that need to manage a pod but:
- Is this correct ? If yes, why does it import pod_health, which seems to me a higher level abstraction to mange the runtime aspect of a pod.
- It mixes method to manipulate the manifest and methods to actually make runtime changes. If so what is its role? Does it abstract a pod manifest or does it expose the lifecycle of a pod? These are two separate concerns which should not be in the same module.

I'd recommend to define which responsibilities each module has and which interface it has, maybe document them either in a README or in docstrings. This will represent your design and the breakdown of your system into logical component which have well defined interfaces to each other.

sentry_streams_k8s/sentry_streams_k8s/operator/pod_health.py
```diff
@@ -0,0 +1,259 @@
...
+def pod_labels(resource: Mapping[str, Any]) -> Mapping[str, str]:
+    return cast(Mapping[str, str], pod_metadata(resource).get("labels", {}) or {})
+
+
+def pod_annotations(resource: Mapping[str, Any]) -> Mapping[str, str]:
+    return cast(Mapping[str, str], pod_metadata(resource).get("annotations", {}) or {})
+
+
+def pod_status(resource: Mapping[str, Any]) -> Mapping[str, Any]:
+    return cast(Mapping[str, Any], resource.get("status", {}) or {})
```
fpacifici: These seem basic access methods for generic fields in a pod, why are in `pod_health` rather than `pod_resource` ?

sentry_streams_k8s/sentry_streams_k8s/operator/pod_health.py
```diff
@@ -0,0 +1,259 @@
...
+    if timestamp is None:
+        return None
+    return (now - timestamp).total_seconds()
+
+
+def pod_is_ready(pod: Mapping[str, Any]) -> bool:
+    status = pod_status(pod)
+    conditions = cast(list[Mapping[str, Any]], status.get("conditions", []) or [])
+    return status.get("phase") == "Running" and any(
+        condition.get("type") == "Ready" and condition.get("status") == "True"
```
fpacifici: Is this the status defined by k8s liveness and readiness probes or is it some internal status you defined ?

sentry_streams_k8s/sentry_streams_k8s/operator/pod_health.py
```diff
@@ -0,0 +1,259 @@
...
+    current_hash = pod_annotations(current).get(SPEC_HASH_ANNOTATION)
+    desired_hash = pod_annotations(desired).get(SPEC_HASH_ANNOTATION)
+    return desired_hash is not None and current_hash != desired_hash
+
+
+def _container_waiting_reason(status: Mapping[str, Any]) -> str | None:
+    state = cast(Mapping[str, Any], status.get("state", {}) or {})
+    waiting = cast(Mapping[str, Any], state.get("waiting", {}) or {})
+    reason = waiting.get("reason")
+    return reason if isinstance(reason, str) else None
```
fpacifici: NOt sure I follow this condition. When can this be non null and also not a string ?

sentry_streams_k8s/sentry_streams_k8s/operator/pod_health.py
```diff
@@ -0,0 +1,259 @@
...
+    """
+    Classify a Pod into a verdict dict.
+
+    Returns keys:
+      name   - the Pod name
+      ready  - True only when Running + Ready condition True
+      reason - short unhealthy reason, or None when healthy
+      delete - whether the reconciler should delete this Pod now
+      force  - whether that deletion should be forced (gracePeriod 0)
+      permanent - True when recovery requires a StreamingPipeline spec update
```
fpacifici: Why does this have to return a dict rather than a properly structured type, dataclass or NAmedTupple ?

Please avoid moving around untyped dicts.

sentry_streams_k8s/sentry_streams_k8s/operator/pod_health.py
```diff
@@ -0,0 +1,259 @@
...
+        status_reason = status.get("reason")
+        status_reason_str = status_reason if isinstance(status_reason, str) else None
+        phase_str = phase if isinstance(phase, str) else None
+        reason = status_reason_str or phase_str or "Terminated"
+        return _verdict(pod_name, reason=reason, delete=True)
+
+    return _verdict(pod_name, ready=pod_is_ready(pod))
+
+
+def pod_status_entry(pod: Mapping[str, Any], health: Mapping[str, Any]) -> dict[str, Any]:
```
fpacifici: What is a pod_status_entry ? 
It seems that this method is tightly coupled with the module that uses it.
It return a concept of status entry: name, ready, phase, ordinal, reason, permanent, that is very much put together ad-hoc for the need of the module that uses it, rather than being a more generic abstraction on the pod. 
This is a smell, it is telling you that this logic should be coupled with the module that uses this method and that, if you need an abstraction over the health of a pod you should provide a generic representation of that. 
I think this is another instance of the concern I raised here: https://github.com/getsentry/streams/pull/345/changes#r3634081525. It seems methods and interface have been added to modules organically instead of being designed deliberately to follow a logical data model.

sentry_streams_k8s/sentry_streams_k8s/operator/reconcile.py
```diff
@@ -0,0 +1,484 @@
...
+        existing = resource.get(name=name, namespace=workload_namespace)
+    except ApiException as e:
+        if e.status != 404:
+            raise
+    else:
+        labels = existing.metadata.labels or {}
+        existing_owner_uid = labels.get(OWNER_UID_LABEL)
+        # Do not take over a resource with the same name from another pipeline:
+        if existing_owner_uid != owner_uid:
+            raise kopf.PermanentError(
```
fpacifici, line 156: Does raising kopf.ParameterError make kopf do anything specific for the operator framework? In other words, why using a framework exception instead of our custom one ?

sentry_streams_k8s/sentry_streams_k8s/operator/reconcile.py
```diff
@@ -0,0 +1,484 @@
...
+) -> None:
+    selector = f"{OWNER_UID_LABEL}={owner_uid}"
+    core = client.CoreV1Api()
+    configmaps = cast(
+        Any,
+        core.list_namespaced_config_map(
+            namespace=workload_namespace,
+            label_selector=selector,
+        ),
+    )
```
fpacifici: Why casting to Any ?
This hsould return a V1ConfigMapList https://k8s-python.readthedocs.io/en/stable/kubernetes.client.models.html?highlight=v1configmaplist#kubernetes.client.models.v1_config_map_list.V1ConfigMapList

Which contains a list of V1ConfigMap objects https://k8s-python.readthedocs.io/en/stable/kubernetes.client.models.html?highlight=v1configmaplist#kubernetes.client.models.v1_config_map.V1ConfigMap

sentry_streams_k8s/sentry_streams_k8s/operator/reconcile.py
```diff
@@ -0,0 +1,484 @@
...
+        )
+
+
+def _allocate_generation(
+    generations: dict[int, int], ordinal: int, pods: list[dict[str, Any]]
+) -> int:
+    live_max = max((pod_generation(pod) for pod in pods), default=-1)
+    generation = max(generations.get(ordinal, -1), live_max) + 1
+
+    if generation > MAX_GENERATION:
```
fpacifici, line 246: Why do we need generations to be ordinal ? Cannot they be arbitrary strings ?
I am thinking whether it is possible to cause a race condition if this method is called twice concurrently. Is this even possible for the same CR ?

sentry_streams_k8s/sentry_streams_k8s/operator/reconcile.py
```diff
@@ -0,0 +1,484 @@
...
+        # TODO: We should see if it is safe to wrap back to generation 0 at this point.
+        raise kopf.PermanentError(f"Replica {ordinal} exceeds {MAX_GENERATION} generations.")
+
+    generations[ordinal] = generation
+    return generation
```
fpacifici, line 251: This looks like a smell. You are returning a new generation, this would make the caller think this method is pure and does not mutate anything, but you also update the generations list in place. IF you want to mutate in place, either return nothing or return the list itself.
If you do not mutate the list then you can return the generation.

sentry_streams_k8s/sentry_streams_k8s/operator/reconcile.py
```diff
@@ -0,0 +1,484 @@
...
+        )
+
+    for pod in current:
+        name = pod_name(pod)
+        health = pod_health(pod, now)
+        health_by_name[name] = health
+        ordinal = pod_ordinal(pod)
+        if ordinal is None or ordinal not in desired_ordinals:
+            _delete_current_pod(dyn, pod, workload_namespace, logger, health, "Stale")
+            continue
```
fpacifici, line 303: Is this meant to happen when we reduce the number of replicas ?

sentry_streams_k8s/sentry_streams_k8s/operator/reconcile.py
```diff
@@ -0,0 +1,484 @@
...
+
+    for ordinal in sorted(desired_ordinals):
+        pods = pods_by_ordinal.get(ordinal, [])
+        desired_template = _build(ordinal, 0)
+        candidates = [
+            pod
+            for pod in pods
+            if (
+                not is_deleting(pod)
+                and not pod_spec_changed(pod, desired_template)
```
fpacifici, line 315: Can a pod spec change within a generation while other pods in the same generation do not change ?

sentry_streams_k8s/sentry_streams_k8s/operator/reconcile.py
```diff
@@ -0,0 +1,484 @@
...
+                and not cast(bool, health_by_name[pod_name(pod)]["delete"])
+            )
+        ]
+
+        if candidates:
+            # Prefer a ready Pod, otherwise keep the newest generation:
+            keep = max(candidates, key=lambda pod: pod_keep_key(pod, health_by_name[pod_name(pod)]))
+            active_pod_names.append(pod_name(keep))
+            generations[ordinal] = max(generations.get(ordinal, -1), pod_generation(keep))
```
fpacifici, line 325: Does this me we would keep a pod of a different generation if unchanged ? I don't think we should entertain the idea of having pods of multiple different generation coexisting outside the rollout and update process.

sentry_streams_k8s/sentry_streams_k8s/operator/reconcile.py
```diff
@@ -0,0 +1,484 @@
...
+    }
+    unhealthy_pods = [entry for entry in pod_statuses if entry.get("reason") is not None]
+    permanent_errors = [entry for entry in unhealthy_pods if entry.get("permanent")]
+    return {
+        "childPods": sorted(active_pod_names),
+        "desiredReplicas": len(desired_ordinals),
+        "readyReplicas": len(ready_ordinals),
+        "unhealthyPods": unhealthy_pods,
+        "permanentErrors": permanent_errors,
+    }
```
fpacifici, line 377: Why returning untyped dictionaries ?

sentry_streams_k8s/sentry_streams_k8s/operator/reconcile.py
```diff
@@ -0,0 +1,484 @@
...
+) -> tuple[dict[str, Any], str]:
+    base_name = deployment["metadata"]["name"]
+
+    if len(base_name) > MAX_BASE_NAME_LENGTH:
+        raise kopf.PermanentError(
+            f"{workload_set} name cannot exceed {MAX_BASE_NAME_LENGTH} characters."
+        )
+
+    template_metadata, template_spec = pod_template_from_deployment(deployment)
+    replicas = deployment["spec"].get("replicas", 0)
```
fpacifici, line 400: Why do we need to deal with deployments here ? The CRD should only manage pods.

sentry_streams_k8s/sentry_streams_k8s/operator/reconcile.py
```diff
@@ -0,0 +1,484 @@
...
+
+    if type(replicas) is not int or replicas < 0:
+        raise kopf.PermanentError(f"{workload_set} replica count must be a non-negative integer.")
+
+    if replicas > MAX_REPLICAS:
+        raise kopf.PermanentError(f"{workload_set} replica count cannot exceed {MAX_REPLICAS}.")
```
fpacifici, line 406: Why imposing this statically ?

sentry_streams_k8s/sentry_streams_k8s/operator/reconcile.py
```diff
@@ -0,0 +1,484 @@
...
+
+    if generation > MAX_GENERATION:
+        # TODO: We should see if it is safe to wrap back to generation 0 at this point.
+        raise kopf.PermanentError(f"Replica {ordinal} exceeds {MAX_GENERATION} generations.")
+
+    generations[ordinal] = generation
+    return generation
+
+
+def reconcile_pipeline_pods(
```
fpacifici, line 254: Why is this function public ? The choice between public and private (ok they are not really private) functions communicates your design. Making it public implies this is the interface of the module, but only another private function in this module calls reconcile_pipeline_pods.

sentry_streams_k8s/sentry_streams_k8s/operator/reconcile.py
```diff
@@ -0,0 +1,484 @@
...
+        owner_namespace=owner_namespace,
+        base_name=base_name,
+        template_metadata=template_metadata,
+        template_spec=template_spec,
+        replicas=replicas,
+        generations=generations,
+        logger=logger,
+        workload_set=workload_set,
+    )
+    if generations != generations_before:
```
fpacifici: Same smell as before. Reading this code I don't know who could have made changes to `generations`.
In order to make reading this code intuitive the data flow should be evident.

sentry_streams_k8s/sentry_streams_k8s/operator/reconcile.py
```diff
@@ -0,0 +1,484 @@
...
+    name: str,
+    namespace: str,
+    uid: str,
+    workload_namespace: str,
+    logger: Logger,
+    patch: kopf.Patch | None = None,
+    status: dict[str, Any] | None = None,
+) -> dict[str, Any]:
+    status_patch = status if status is not None else patch.status if patch is not None else None
+    consumer = from_crd_spec(dict(spec), name=name)
```
fpacifici, line 484: What's the type of spec here ?

sentry_streams_k8s/sentry_streams_k8s/operator/reconcile.py
```diff
@@ -0,0 +1,484 @@
...
+    try:
+        validate(consumer)
+        result = render(consumer)
```
fpacifici, line 487: IS this rendering the deployment or a pod ?

sentry_streams_k8s/sentry_streams_k8s/operator/reconcile.py
```diff
@@ -0,0 +1,484 @@
...
+            deployment=deployment,
+            workload_set=workload_set,
+            workload_namespace=workload_namespace,
+            owner_uid=uid,
+            owner_name=name,
+            owner_namespace=namespace,
+            logger=logger,
+        )
+        pod_set_results[workload_set] = set_result
+        ledger_names.add(ledger_name)
```
fpacifici: Not sure I follow this process. Why did we render the deployment to create the pods ? 
If this operator manages the pods directly there is no reason to generate the deployment. Or is it just meant to speed up the iteration ?

sentry_streams_k8s/sentry_streams_k8s/operator/reconcile.py
```diff
@@ -0,0 +1,484 @@
...
+    configmap = result["configmap"]
+
+    _prepare_manifest(
+        configmap,
+        workload_namespace=workload_namespace,
+        owner_uid=uid,
+        owner_name=name,
+        owner_namespace=namespace,
+    )
+    _apply(dyn, configmap, workload_namespace=workload_namespace, owner_uid=uid)
```
fpacifici: Is this apply going to trigger the rollout of the pods as, if the config change, the pods have to be restarted even if their manifest does not change.

sentry_streams_k8s/sentry_streams_k8s/operator/reconcile.py
```diff
@@ -0,0 +1,484 @@
...
+
+    rendered_sets = {PRIMARY_WORKLOAD_SET: result["deployment"]}
+    if "canary_deployment" in result:
+        rendered_sets[CANARY_WORKLOAD_SET] = result["canary_deployment"]
+
+    pod_set_results: dict[str, dict[str, Any]] = {}
+    ledger_names: set[str] = set()
+    for workload_set, deployment in rendered_sets.items():
```
fpacifici, line 511: Do you plan to support canary ?

sentry_streams_k8s/sentry_streams_k8s/operator/operator.py
```diff
@@ -1,30 +1,74 @@
...
+
+
+class ReconcileScheduler:
+    def __init__(self, max_concurrent: int = MAX_CONCURRENT_RECONCILES) -> None:
+        self._events: dict[str, asyncio.Event] = {}
+        self._locks: dict[str, asyncio.Lock] = {}
+        self.limit = asyncio.Semaphore(max_concurrent)
+
+    def register(self, uid: str) -> asyncio.Event:
+        event = self._events.setdefault(uid, asyncio.Event())
```
fpacifici, line 45: What happens if you try to re-register the same uid ?

sentry_streams_k8s/sentry_streams_k8s/operator/operator.py
```diff
@@ -1,30 +1,74 @@
...
+        return event
+
+    def unregister(self, uid: str, event: asyncio.Event) -> None:
+        if self._events.get(uid) is event:
+            self._events.pop(uid, None)
+
+    def notify(self, uid: str) -> bool:
+        event = self._events.get(uid)
+        if event is None:
+            return False
```
fpacifici, line 56: When could this happen ?

sentry_streams_k8s/sentry_streams_k8s/operator/operator.py
```diff
@@ -34,188 +78,174 @@ def _workload_namespace() -> str:
...
     **_: Any,
 ) -> None:
-    assert namespace is not None
-    workload_namespace = _workload_namespace()
+    if namespace is None:
+        raise kopf.PermanentError("Missing namespace!")
 
-    consumer = from_crd_spec(dict(spec), name=name)
+    scheduler = _scheduler(memo)
+    event = scheduler.register(uid)
```
fpacifici, line 215: Does this happen once per CR or once where we create the CRD ?

sentry_streams_k8s/sentry_streams_k8s/operator/operator.py
```diff
@@ -34,188 +78,174 @@ def _workload_namespace() -> str:
...
+    logger: Logger,
+    **_: Any,
+) -> None:
+    if type not in {"DELETED", "MODIFIED"}:
+        return
+
+    if type == "MODIFIED" and meta.get("deletionTimestamp") is None:
+        health = pod_health(dict(body), datetime.now(timezone.utc))
+        if not health["delete"]:
+            return
```
fpacifici: When can this happen? you have deletionTimestamp but health["delete"] be False ?

sentry_streams_k8s/sentry_streams_k8s/operator/pod_health.py
fpacifici, line 1: One general comment about the typing and design of the modules that manage pods and deployment specs:
Your design is procedural, standalone functions that transform input structures into output, there is nothing inherently wrong in this though the issue is that you are transforming loosely typed dictionaries rather than more precisely typed objects.

This makes your code very verbose as you have to add all those helper methods to extract specific fields from a nested dictionary and very fragile as it is easy to make mistakes like adding keys or reading the wrong keys.

Please let's have a properly typed data structure to manipulate the Kubernetes resources specs:
- The CoreAPI provides strongly typed objects to manage any resource: See here https://k8s-python.readthedocs.io/en/stable/kubernetes.client.models.html?highlight=v1configmaplist#kubernetes.client.models.v1_config_map.V1ConfigMap. These remove the need for most of the helper methods. 
- If there is a problem with those, Please create at least your own TypedDict or Dataclasses to represent the data structures. 
- If you need to use the DynamicClient Then turn what you get from the API in a typed datastructure right after calling the API and transform the typed data structure to dict right before calling the apply/patch api. By the way, why do you need to use the dynamic client ? 

We should remove all the management of Dict[str, Any]

sentry_streams_k8s/sentry_streams_k8s/operator/pod_resources.py
```diff
@@ -0,0 +1,200 @@
...
+    return (cast(bool, health["ready"]), pod_generation(pod), pod_name(pod))
+
+
+def pod_template_from_deployment(
+    deployment: Mapping[str, Any],
+) -> tuple[Mapping[str, Any], Mapping[str, Any]]:
+    template = cast(Mapping[str, Any], deployment["spec"]["template"])
+    return cast(Mapping[str, Any], template.get("metadata", {}) or {}), cast(
+        Mapping[str, Any], template["spec"]
+    )
```
fpacifici: A general note about this procedural design. If you feel this is the best option I would not push back as this is your coding style and it is a legit design. Though, since your system is doing a lot of manipulation of complex nested data structures, have you considered making this more object oriented by creating classes for the main concepts you are manipulating (Spec, Pod, ConfigMap, etc) and add methods to extract or transform them, plus modules with the application logic to reconcile and call the api ? 
This would make a design emerge as it would:
- force you to package the methods with the data types they manipulate 
- separate the data type manipulation from the execution of the application logic (calling api) 

IF you like sticking with this style better I think you will have to do more work to properly modularize the system:
- You will need properly structured types rather than relying on untyped dictionaries 
- You need to make the break down into modules more deliberate: examples could be having a `pod` module, a `configmap` module, maybe `container` or `status`. Each one of these contains the functions to manipulate these data structures. Then you'll need modules to perform the orchestration work: apply/delete/patch etc.

sentry_streams_k8s/tests/test_pod_health.py
fpacifici, line 1: This test is only testing the `pod_health` function.
Please add one single test to test the various accessor functions that return data from the pod or make them private (`pod_metadata`, `pod_labels`, etc)

sentry_streams_k8s/sentry_streams_k8s/operator/pod_resources.py
fpacifici, line 1: I do not see unit tests for this module. Am I missing something ?

sentry_streams_k8s/tests/test_pods.py
```diff
@@ -0,0 +1,420 @@
...
+        list_pods,
+    )
+    monkeypatch.setattr(
+        "sentry_streams_k8s.operator.reconcile.apply_pod",
+        lambda _dyn, pod, _namespace: applied.append(dict(pod)),
+    )
+    monkeypatch.setattr(
+        "sentry_streams_k8s.operator.reconcile.delete_pod",
+        lambda _dyn, name, _namespace, force=False: deleted.append((name, force)),
+    )
```
fpacifici, line 129: Please do not monkeypatch methods you built.
This test seems like an integration test. So you should avoid mocking or monkeypatching what is not strictly needed. Please mock just the k8s API object.

Also I think this is monkeypatching the wrong path as delete_pod, apply_pod and list_owned_pods are not in the reconcile module. Unless you need to monkeypatch them there because that's where they are used.

sentry_streams_k8s/tests/test_pods.py
```diff
@@ -0,0 +1,420 @@
...
+        template_spec=template_spec or spec,
+        replicas=replicas,
+        generations=ledger,
+        logger=MagicMock(),
+        workload_set=workload_set,
+    )
+    return applied, deleted, result, ledger
+
+
+def test_build_pipeline_pod_stamps_identity_without_mutating_template() -> None:
```
fpacifici, line 149: What does this mean ?  (I guess this is AI generated, am I right ? )

fpacifici (changes requested):
Please see my comments inline for the details. Overall:
- I don't think you need a configMap for the generaiton state. If you can do that with the status of a CR, please avoid the configMap complexity.
- Some high level design constraints: Untyped dictionaries, breakdown into modules, public/private interfaces. Comments are inline
- Consider breaking this down to make the task of reviewing tractable. A suggestion: One PR for the pod spec manipulation logic, one PR for the reconcile code, one PR that wire things up together in the operator. Failing to do so will make the review process really long as it requires the reviewer to allocate a lot of time to review each iteration.

sentry_streams_k8s/sentry_streams_k8s/operator/reconcile.py
bmcquilkin-sentry: PermanentError means that kopf will no longer try to reconcile the CR until it is updated since the error is unrecoverable (e.g. bad image or something improperly set up - the operator should not retry for these since no matter how many times it retries it will still be broken until the CR is updated). TemporaryError is when the error can be recoverable without change and the operator should try again (e.g. network error while fetching the image - the operator should retry here since it is not a problem with the CR).

sentry_streams_k8s/sentry_streams_k8s/operator/operator.py
bmcquilkin-sentry: Each CR gets its own reconcile daemon.

sentry_streams_k8s/sentry_streams_k8s/operator/reconcile.py
bmcquilkin-sentry: Canary deployment is already supported as its own set.

fpacifici: Could you expand please? 
How does the pipeline CRD start a canary consumer if the number of replicas is higher than 1 and canary flag is set ?

sentry_streams_k8s/sentry_streams_k8s/operator/reconcile.py
fpacifici: Please add a comment explaining why you use that exception.

sentry_streams_k8s/sentry_streams_k8s/operator/reconcile.py
bmcquilkin-sentry: This should also answer your question about what a workload set is. A workload set is a just a set of Pods. There is one for primary and one for canary. Replicas are in the primary set, and the canary Pod is in the canary set. When we change replicas, it adjusts from the primary set. When we turn canary on/off, it creates or deletes the canary set

PR #346, opened 2026-07-22
feat(operator): static membership

bmcquilkin-sentry:
Adds static membership to consumers (and also adds some type fix mypy was complaining about).


evanh (approved):

sentry_streams/sentry_streams/adapters/arroyo/rust_arroyo.py
```diff
@@ -144,20 +145,29 @@ def build_initial_offset(offset_reset: str) -> InitialOffset:
...
     bootstrap_servers = consumer_config["bootstrap_servers"]
     group_id = (
         consumer_group_override or consumer_config.get("consumer_group") or f"pipeline-{source}"
     )
     auto_offset_reset = build_initial_offset(consumer_config.get("auto_offset_reset", "latest"))
     strict_offset_reset = bool(consumer_config.get("strict_offset_reset", False))
-    override_params = consumer_config.get("override_params", {})
+    override_params = dict(consumer_config.get("override_params", {}))
+
+    group_instance_id = os.environ.get("STREAMS_KAFKA_GROUP_INSTANCE_ID")
```
fpacifici: You do not need to read environment variables from inside the adapter. The code that reads the config file is already able to transparently swap a config field with an environment variable provided according to a standard format 
https://github.com/getsentry/streams/blob/main/sentry_streams/sentry_streams/pipeline/config.py#L26-L32

See an example of how this is used 
- https://github.com/getsentry/streams/blob/main/sentry_streams/tests/pipeline/test_config.py#L45-L57
- https://github.com/getsentry/ops/blob/master/k8s/services/super-big-consumers/_items_log_config.yaml#L10

This means you can leverage the standard override mechanism without doing anything special for group.instance.id here.

More in general, beware of reading random environment variables deep in the stack. That makes it very hard to track down how to configure your system. 
A better option is to have a config abstraction that provide the whole configuration and takes care to resolve, config files, overrides with environment variables and other ways to configure the system.
Then you separate the concerns cleanly. One system manages the loading of the config and another consumes it. The system consuming the config does not have to care where the value comes from.

sentry_streams/tests/adapters/arroyo/test_rust_arroyo.py
```diff
@@ -46,6 +46,7 @@ def test_rust_arroyo_adapter(
     [
         pytest.param(
             {
+                "starts_segment": None,
```
fpacifici: Why this ? `start_segment` is not an attribute of KafkaConsumerConfig

sentry_streams/tests/adapters/arroyo/test_rust_arroyo.py
```diff
@@ -157,3 +160,54 @@ def test_build_kafka_producer_config(
     assert result is not None
     assert list(result.bootstrap_servers) == expected_bootstrap_servers
     assert result.override_params == expected_override_params
+
+
+def test_build_kafka_consumer_config_injects_static_membership(
+    monkeypatch: pytest.MonkeyPatch,
+) -> None:
+    monkeypatch.setenv("STREAMS_KAFKA_GROUP_INSTANCE_ID", "consumer-2")
```
fpacifici: You do not need to monkeypatch environment variables. Just set that with `os.environ["blabla"] = "blabla"`.

In general please try to avoid monkey patching whenever it is not strictly necessary. It is really hard to makes sense of the behavior of the class you are patching when it evolves.

fpacifici (approved):
Pelase address my comments on the usage of hte environment variable.

fpacifici (changes requested):
Clicked the wrong button, please see my comments above

sentry_streams/tests/adapters/arroyo/test_rust_arroyo.py
bmcquilkin-sentry: [KafkaConsumerConfig](https://github.com/getsentry/streams/blob/cfec6679df3b5d312324ade1b8e88a42ab5479c4/sentry_streams/sentry_streams/config_types.py#L23) extends StepConfig which contains `starts_segment`

sentry_streams/sentry_streams/adapters/arroyo/rust_arroyo.py
bmcquilkin-sentry: If I am understanding right: Config specifies `group.instance.id: "${envvar:KAFKA_GROUP_INSTANCE_ID}"` in override params and the operator sets the `KAFKA_GROUP_INSTANCE_ID` variable to the Pod identity?

If this is the case, is there a reason why we wouldn't just want the operator to autofill `group.instance.id` in override params to avoid requiring the configs to specify it? I can't think of a use case where we would ever want to override it manually except for disabling static membership

fpacifici: > If I am understanding right: Config specifies group.instance.id: "${envvar:KAFKA_GROUP_INSTANCE_ID}" in override params and the operator sets the KAFKA_GROUP_INSTANCE_ID variable to the Pod identity?

Yes

> If this is the case, is there a reason why we wouldn't just want the operator to autofill group.instance.id in override params to avoid requiring the configs to specify it? I can't think of a use case where we would ever want to override it manually except for disabling static membership

This is a good point. I don't see the solution I suggested as ideal. I think we still miss a piece of infrastructure to allow the operator to talk to the pod. Though I think the custom approach to read the envvar is worse. The config can still provide instance id and this can be overridden. This override is quite arbitrary as it is specific of the operator.

The clean way to do this would be one of these I believe:
- Make the operator provide the instance id via the HTTP server when triggering the subscription operation. So this is constrained to the operator management.
- If we had a config override standard mechanism the operator could add an override file to mount to override the main config.
- Implement a general way to force the override of config per pod. This would likely be via envvar or by mounting an override file. Though We would manage it in the config.py module.

If you want you can override the instance id with the envvar inside the config.py module bypassing the content of the config file. That I would be ok with.

PR #352, opened 2026-07-25
feat(arroyo): allow external shutdowns

bmcquilkin-sentry:
Implements the shutdown method in rust_arroyo for when the operator will need to shutdown manually for blue-green deployment.

### DIFF
--- a/sentry_streams/src/consumer.rs
+++ b/sentry_streams/src/consumer.rs
@@ -31,7 +31,7 @@ use sentry_arroyo::processing::ProcessorHandle;
 use sentry_arroyo::processing::StreamProcessor;
 use sentry_arroyo::types::{Message, Topic};
 use std::collections::HashMap;
-use std::sync::Arc;
+use std::sync::{Arc, Mutex};
 
 /// Default path for the healthcheck file touched when write_healthcheck is enabled.
 /// Matches Arroyo docs for Kubernetes liveness probes.
@@ -87,7 +87,7 @@ pub struct ArroyoConsumer {
 
     /// The ProcessorHandle allows the main thread to stop the StreamingProcessor
     /// from a different thread.
-    handle: Option<ProcessorHandle>,
+    handle: Mutex<Option<ProcessorHandle>>,
 
     // this variable must live for the lifetime of the entire consumer.
     // This is a requirement of Arroyo Rust.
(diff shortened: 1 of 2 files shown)


sentry_streams/src/consumer.rs
```diff
@@ -162,8 +162,8 @@ impl ArroyoConsumer {
 
     /// Runs the consumer.
     /// This method is blocking and will run until the consumer
-    /// is stopped via SIGTERM or SIGINT.
-    fn run(&mut self) {
```
fpacifici, line 166: Why do we need to pass the python interpreter here ?

sentry_streams/src/consumer.rs
```diff
@@ -205,16 +208,19 @@ impl ArroyoConsumer {
...
-            sentry::capture_error(&e);
-        }
+        py.detach(|| {
+            if let Err(e) = processor.run() {
+                tracing::error!("StreamProcessor error: {:?}", e);
+                sentry::capture_error(&e);
+            }
+        });
+
+        *self.handle.lock().unwrap() = None;
```
fpacifici, line 218: I am not sure I get why we need to add a mutex here. Specifically:
1. The get_handle method on the `processor` struct makes a copy of the handle https://github.com/getsentry/arroyo/blob/main/rust-arroyo/src/processing/mod.rs#L471-L473. The method does not require a mutable reference. So if you need to control the shutdown from two places, you should even be able to keep two copies of the handle, one managed by shutdown and one managed by the signal handler.
2. Even if you kept a single copy of it, why do you need to put the `processor.run()` call in a `py` block ? And Why do you need to add a mutex ?

fpacifici (changes requested):
Please see my comment, I don't think we should put the run method in a py block and we should not need a mutex.

sentry_streams/src/consumer.rs
bmcquilkin-sentry: `processor.run` is blocking so we need py.deatch to release the GIL, since the server in Python still needs to be able to run and process events while the consumer is running.

bmcquilkin-sentry: For the mutex, I believe it is required for the handle to be properly shared between run and shutdown?

PR #353, opened 2026-07-25
feat(streams): operator control server

bmcquilkin-sentry:
Basic control server for the operator to use for blue-green deployments (readyz, status start, stop). Adds --control-port option to runner to determine if the process should use deployment mode (autostart) or operator mode (wait for /start).

### DIFF
--- a/sentry_streams/sentry_streams/adapters/stream_adapter.py
+++ b/sentry_streams/sentry_streams/adapters/stream_adapter.py
@@ -44,6 +44,12 @@ StreamT = TypeVar("StreamT")
 StreamSinkT = TypeVar("StreamSinkT")
 
 
+class RuntimeStateError(RuntimeError):
+    """
+    Raised when an operation conflicts with the runtime's lifecycle state.
+    """
+
+
 class RuntimeState(StrEnum):
     IDLE = "idle"
     STARTING = "starting"
@@ -56,32 +62,36 @@ class RuntimeState(StrEnum):
     def is_terminal(self) -> bool:
         return self in (RuntimeState.STOPPED, RuntimeState.ERRORED)
 
+    @property
+    def rank(self) -> int:
+        if self is RuntimeState.IDLE:
+            return 0
+        if self is RuntimeState.STARTING:
+            return 1
+        if self is RuntimeState.CONSUMING:
+            return 2
+        if self is RuntimeState.STOPPING:
+            return 3
+        return 4
 
-# To simplify lifecycle and state transitions we can use a forward-only ranking.
-# A higher-ranked state should realistically never transition to a lower one.
-# Would need to be updated if we allow restarting (but we don't for now).
-
-_STATE_RANKS: Mapping[RuntimeState, int] = {
-    RuntimeState.IDLE: 0,
-    RuntimeState.STARTING: 1,
-    RuntimeState.CONSUMING: 2,
-    RuntimeState.STOPPING: 3,
-    RuntimeState.STOPPED: 4,
-    RuntimeState.ERRORED: 4,
-}
+    def can_transition_to(self, state: RuntimeState) -> bool:
+        return self.rank < state.rank
 
 
 @dataclass(frozen=True)
 class RuntimeStatus:
     state: RuntimeState
-    error: str | None = None
+    error: Exception | None = None
 
     @property
     def is_terminal(self) -> bool:
         return self.state.is_terminal
 
     def as_dict(self) -> dict[str, str | None]:
-        return {"state": self.state.value, "error": self.error}
+        return {
+            "state": self.state.value,
+            "error": str(self.error) if self.error is not None else None,
+        }
 
 
 class StreamAdapter(ABC, Generic[StreamT, StreamSinkT]):
@@ -100,40 +110,49 @@ class StreamAdapter(ABC, Generic[StreamT, StreamSinkT]):
         with self.__status_lock:
             return self.__status
 
-    def _set_status(self, state: RuntimeState, error: str | None = None) -> None:
+    def _set_status(self, state: RuntimeState, error: Exception | None = None) -> None:
         with self.__status_lock:
-            if _STATE_RANKS[state] <= _STATE_RANKS[self.__status.state]:
+            if not self.__status.state.can_transition_to(state):
                 return
             self.__status = RuntimeStatus(state, error)
 
     def begin_start(self) -> RuntimeStatus:
         with self.__status_lock:
-            if self.__status.state is RuntimeState.IDLE:
-                self.__status = RuntimeStatus(RuntimeState.STARTING)
+            state = self.__status.state
+            if state.is_terminal:
+                raise RuntimeStateError(f"cannot restart runtime that is {state}")
+            if state is not RuntimeState.IDLE:
+                raise RuntimeStateError(f"cannot start runtime while it is {state}")
+            self.__status = RuntimeStatus(RuntimeState.STARTING)
             return self.__status
 
     def run(self) -> None:
-        if self.begin_start().state is not RuntimeState.STARTING:
+        state = self.status.state
+        if state is RuntimeState.STOPPING:
             self._set_status(RuntimeState.STOPPED)
             return
+        if state is not RuntimeState.STARTING:
+            raise RuntimeStateError(f"cannot run runtime while it is {state}")
 
         try:
             self._run()
-        except BaseException as exc:
-            self._set_status(RuntimeState.ERRORED, str(exc))
+        except Exception as exc:
+            self._set_status(RuntimeState.ERRORED, exc)
             raise
         else:
             self._set_status(RuntimeState.STOPPED)
 
-    def shutdown(self) -> None:
+    def shutdown(self) -> RuntimeStatus:
         with self.__status_lock:
             state = self.__status.state
-            if not state.is_terminal:
-                self.__status = RuntimeStatus(
-                    RuntimeState.STOPPED if state is RuntimeState.IDLE else RuntimeState.STOPPING
-                )
+            if state is RuntimeState.STOPPING or state.is_terminal:
+                return self.__status
+            self.__status = RuntimeStatus(
+                RuntimeState.STOPPED if state is RuntimeState.IDLE else RuntimeState.STOPPING
+            )
 
         self._shutdown()
+        return self.status
 
     @classmethod
     @abstractmethod
--- a/sentry_streams/sentry_streams/adapters/stream_adapter.py
+++ b/sentry_streams/sentry_streams/adapters/stream_adapter.py
@@ -1,6 +1,9 @@
 from __future__ import annotations
 
+import threading
 from abc import ABC, abstractmethod
+from dataclasses import dataclass
+from enum import StrEnum
 from typing import (
     Any,
     Callable,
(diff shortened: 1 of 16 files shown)


sentry_streams/sentry_streams/control_server.py
fpacifici, line 1: Please put this in its own module

sentry_streams/sentry_streams/control_server.py
```diff
@@ -0,0 +1,158 @@
+"""
+Operator control server for a consumer process.
+
+When the consumer runs under the operator's blue-green deployment it cannot begin
+consuming as soon as it starts. Instead is should prepare and then report itself
+ready, then wait for the operator to explicitly start it when handing off.
```
fpacifici: It is not only useful for that. 
A control HTTP endpoint can be used for many reasons including capturing telemetry

sentry_streams/sentry_streams/control_server.py
```diff
@@ -0,0 +1,158 @@
...
+STATE_IDLE = "idle"
+STATE_STARTING = "starting"
+STATE_CONSUMING = "consuming"
+STATE_STOPPING = "stopping"
+STATE_STOPPED = "stopped"
+STATE_ERRORED = "errored"
+
+
+class ConsumerController:
+    def __init__(self, runtime: Any) -> None:
```
fpacifici: Why is runtime Any ?

sentry_streams/sentry_streams/control_server.py
```diff
@@ -0,0 +1,158 @@
...
+            except Exception as exc:
+                logger.exception("control-server request failed: %s", self.path)
+                self._respond(500, {"error": str(exc)})
+
+    return ControlHandler
+
+
+def serve(runtime: Any, port: int) -> None:
+    controller = ConsumerController(runtime)
+    server = ThreadingHTTPServer(("0.0.0.0", port), _make_handler(controller))
```
fpacifici: Please allow the user to customize the bind host as well.

sentry_streams/sentry_streams/control_server.py
```diff
@@ -0,0 +1,158 @@
...
+STATE_STOPPING = "stopping"
+STATE_STOPPED = "stopped"
+STATE_ERRORED = "errored"
+
+
+class ConsumerController:
+    def __init__(self, runtime: Any) -> None:
+        self._runtime = runtime
+        self._lock = threading.Lock()
+        self._state = STATE_IDLE
```
fpacifici: Could you expand on why you need to keep the StreamingAdapter state inside the webserver? It seems it would be much simpler to make the StreamingAdapter expose its state. It has a lot more context on what it is going on and when it is starting/stopping/failing/etc.

sentry_streams/sentry_streams/control_server.py
```diff
@@ -0,0 +1,158 @@
...
+            self._error = None
+            self._thread = threading.Thread(target=self._run, name="consumer-run", daemon=False)
+            self._thread.start()
+            return self._state
+
+    def _run(self) -> None:
+        with self._lock:
+            self._state = STATE_CONSUMING
+        try:
+            self._runtime.run()
```
fpacifici: This is a control server that is just supposed to manage a running pipeline. Now it seems that the adapter is owned entirely by the control server, which means we cannot have two separate mechanisms to control the adapter.

I would recommend this:
1. The main thread starts the control system that can receive commands by different interfaces, whether that is intercepting a signal or an http endpoint or others. 
2. The HTTP server gets a reference to such controller and sends commands on that object that immediately return. 

Now the question would be whether the adapter itself is supposed to be the control object that starts, hangs and has other methods to be controller, or whether there has to be a separate Controller to manage the life cycle in a thread, this controlelr would call `run`. I think both would be valid options, though the controller should be instantiated by the main runner and not by the httpserver.

Do SIGINT and SIGTERM even work this way ?

sentry_streams/sentry_streams/adapters/stream_adapter.py
```diff
@@ -41,13 +44,97 @@
...
+        self.__status = RuntimeStatus(RuntimeState.IDLE)
+
+    @property
+    def status(self) -> RuntimeStatus:
+        with self.__status_lock:
+            return self.__status
+
+    def _set_status(self, state: RuntimeState, error: str | None = None) -> None:
+        with self.__status_lock:
+            if _STATE_RANKS[state] <= _STATE_RANKS[self.__status.state]:
```
fpacifici: Please move the ranking logic inside the RuntimeState enum. Enums can have method.

sentry_streams/sentry_streams/adapters/stream_adapter.py
```diff
@@ -41,13 +44,97 @@
...
+                return
+            self.__status = RuntimeStatus(state, error)
+
+    def begin_start(self) -> RuntimeStatus:
+        with self.__status_lock:
+            if self.__status.state is RuntimeState.IDLE:
```
fpacifici: Shouldn't this fail if state is not Idle ?
Also what is this state for ?

sentry_streams/sentry_streams/adapters/stream_adapter.py
```diff
@@ -41,13 +44,97 @@
...
+                self.__status = RuntimeStatus(RuntimeState.STARTING)
+            return self.__status
+
+    def run(self) -> None:
+        if self.begin_start().state is not RuntimeState.STARTING:
+            self._set_status(RuntimeState.STOPPED)
+            return
```
fpacifici: Not following this logic. Why would we put this in stopped state if the state is not starting ?

sentry_streams/sentry_streams/adapters/stream_adapter.py
```diff
@@ -41,13 +44,97 @@
...
+
+        try:
+            self._run()
+        except BaseException as exc:
```
fpacifici: Why BaseException instead of any Exception ?

sentry_streams/sentry_streams/adapters/stream_adapter.py
```diff
@@ -41,13 +44,97 @@
...
+            self._set_status(RuntimeState.ERRORED, str(exc))
```
fpacifici: Please store the exception object and not the string only otherwise you loose the stacktrace.

sentry_streams/sentry_streams/server/control_server.py
```diff
@@ -0,0 +1,97 @@
+"""
+Operator control server for a consumer process.
```
fpacifici: This is not enough.  It does not add any information to the previous state where there was no docstring.
Explain:
- What a control server for a consumer process is 
- What it does 
- Which kind of operations it supports
- Why do we have an HTTP server.

Also, please move it as docstring of the ControlServer class.

sentry_streams/sentry_streams/control.py
```diff
@@ -0,0 +1,104 @@
...
+    """
+
+
+class PipelineController:
+    """
+    Control panel for one streaming pipeline. Starts the adapter's blocking run loop
+    in a background thread and allows concurrent requests to start/stop the pipeline.
+
+    A pipeline is single-use: after it stops or fails, it cannot be restarted.
+    A replacement deployment must create a new controller and adapter.
```
fpacifici, line 19: Please explain why you need a lock.

sentry_streams/sentry_streams/control.py
```diff
@@ -0,0 +1,104 @@
...
+    def request_start(self) -> RuntimeStatus:
+        """
+        Ask the pipeline to start (non-blocking).
+        """
+        with self._lock:
+            state = self._runtime.status.state
+            if state is RuntimeState.STOPPING:
+                raise PipelineStateError("cannot start a pipeline that is stopping")
+            if state.is_terminal:
+                raise PipelineStateError(f"cannot restart a pipeline that is {state}")
```
fpacifici: This seems something you could put in the adapter that manages the valid state transitions. splitting the logic that manages state transitions between classes make the state machine hard to follow.

sentry_streams/sentry_streams/control.py
```diff
@@ -0,0 +1,104 @@
...
+        """
+        with self._lock:
+            status = self._runtime.status
+
+            if status.state is RuntimeState.STOPPING or status.is_terminal:
+                return status
+
+            self._runtime.shutdown()
+            if self._thread is None:
+                self._finished.set()
```
fpacifici: How can this happen ?

sentry_streams/sentry_streams/control.py
```diff
@@ -0,0 +1,104 @@
...
+
+            return self._runtime.status
+
+    def wait_until_finished(self, timeout: float | None = None) -> RuntimeStatus:
+        """
+        Wait until this controller is completely finished.
+        """
+        self._finished.wait(timeout)
```
fpacifici: Isn't this the same as the termination of the thread ?

sentry_streams/sentry_streams/runner.py
```diff
@@ -23,9 +31,95 @@
...
+def _install_signal_handlers(shutdown_requested: threading.Event) -> None:
+    """Turn SIGINT/SIGTERM into a shutdown request."""
+
+    def _handle_termination(signum: int, _frame: FrameType | None) -> None:
+        logger.info("received signal %d; requesting pipeline shutdown", signum)
+        shutdown_requested.set()
+
+    for signum in (signal.SIGINT, signal.SIGTERM):
+        signal.signal(signum, _handle_termination)
+
```
fpacifici, line 51: Does this work with the rust adapter ?
https://github.com/getsentry/streams/blob/main/sentry_streams/src/consumer.rs#L202-L206

We had to add the handler in the rust code before.

sentry_streams/sentry_streams/runner.py
```diff
@@ -23,9 +31,95 @@
...
+            serve_thread.join(SHUTDOWN_TIMEOUT_SEC)
+
+    if serve_failure:
+        raise serve_failure[0]
+
+    return controller.snapshot
+
+
+def run_runtime(runtime: StreamAdapter[Any, Any]) -> None:
+    shutdown_requested = threading.Event()
```
fpacifici, line 120: I don't think you need this.
You can pass the signal handler a partial function with a reference to the controller so that the handler can directly invoke the shutdown method on the controller.

sentry_streams/sentry_streams/control.py
```diff
@@ -0,0 +1,104 @@
...
+
+        return self._runtime.status
+
+    def _run_runtime(self) -> None:
+        try:
+            self._runtime.run()
+        except Exception:
+            logger.exception("pipeline run loop failed")
+        finally:
+            self._finished.set()
```
fpacifici: This is not ok, we will loose the exception stacktrace.
Please let the thread bubble up the exception and capture it with this https://docs.python.org/3/library/threading.html#threading.excepthook.

In alternative you can use a Future to have a higher level abstraction where you can start a thread and get the exception just by capping `result` 
https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.Executor.submit

sentry_streams/sentry_streams/runner.py
```diff
@@ -23,9 +31,95 @@
...
+        serve_thread.start()
+    else:
+        controller.request_start()
+
+    try:
+        controller.wait_until_finished()
+    finally:
+        shutdown_requested.set()
+        shutdown_thread.join(SHUTDOWN_TIMEOUT_SEC)
+        controller.request_stop()
```
fpacifici: Why do you need to request_stop on the controller that has already finished ?

sentry_streams/src/consumer.rs
```diff
@@ -85,9 +87,13 @@ pub struct ArroyoConsumer {
...
     steps: Vec<Py<RuntimeOperator>>,
 
+    /// Set by `shutdown()` so a stop that arrives before `run()` builds the
+    /// processor is not lost. `run()` checks it before entering the run loop.
+    shutdown_requested: AtomicBool,
+
     /// The ProcessorHandle allows the main thread to stop the StreamingProcessor
     /// from a different thread.
-    handle: Option<ProcessorHandle>,
+    handle: Mutex<Option<ProcessorHandle>>,
```
fpacifici, line 96: Why do you need a Mutex here ?

sentry_streams/src/consumer.rs
```diff
@@ -196,25 +204,40 @@ impl ArroyoConsumer {
...
+            processor.get_handle().signal_shutdown();
+        }
+
+        // The GIL is released around the run loop so that a Python control
+        // thread can call shutdown while this consumer is running:
 
-        if let Err(e) = processor.run() {
-            tracing::error!("StreamProcessor error: {:?}", e);
-            sentry::capture_error(&e);
+        let run_error = py.detach(|| match processor.run() {
```
fpacifici, line 217: So here is the issue. processor.run() takes the GIL in several places. Did you see this code running successfully with a real pipeline that runs python code ?

sentry_streams/src/consumer.rs
```diff
@@ -196,25 +204,40 @@ impl ArroyoConsumer {
...
-        ctrlc::set_handler(move || {
-            println!("\nCtrl+C pressed!");
-            handle.signal_shutdown();
-        })
-        .expect("Error setting Ctrl+C handler");
+        *self.handle.lock().unwrap() = Some(processor.get_handle());
+
+        if self.shutdown_requested.load(Ordering::Acquire) {
+            processor.get_handle().signal_shutdown();
+        }
```
fpacifici, line 212: I don;'t think you need this. You can get multiple handles from the processor. You do not need to share it between run and shutdown.

fpacifici (changes requested):
Please see the comments in line

sentry_streams/src/consumer.rs
bmcquilkin-sentry: The run call has a reference to the processor on one thread, shutdown does not have a reference (processor is a local to run) and is on another thread. Run needs to use a mutex to store the handle so that shutdown can access it.

sentry_streams/src/consumer.rs
bmcquilkin-sentry: See: https://github.com/getsentry/streams/pull/353#discussion_r3692696933

sentry_streams/src/consumer.rs
bmcquilkin-sentry: Yes, we need detach because otherwise the control server in Python would be locked and not respond.

sentry_streams/sentry_streams/runner.py
bmcquilkin-sentry: Yes because before the run was blocking Python (GIL was locked) so a Python signal handler couldn't run, but now since it runs separately and is detached it can process and shutdown the adapter.

sentry_streams/sentry_streams/control.py
bmcquilkin-sentry: If a stop request arrives before the worker thread starts

sentry_streams/sentry_streams/control.py
bmcquilkin-sentry: Technically it can also get triggered if the thread never runs (stopped while idle)

PR #355, opened 2026-07-29
ref(operator): extract reconcile and use generated clients

bmcquilkin-sentry:
Avoid using DynamicClient and use the SSA patching with typed clients instead. Extracts reconcile behavior to separate modules to be expanded later.

### DIFF
--- a/sentry_streams_k8s/sentry_streams_k8s/operator/constants.py
+++ b/sentry_streams_k8s/sentry_streams_k8s/operator/constants.py
@@ -7,6 +7,11 @@ PLURAL = "streamingpipelines"
 FIELD_MANAGER = "streaming-operator"
 WORKLOAD_NAMESPACE_ENV = "WORKLOAD_NAMESPACE"
 
+# SSA through API clients is tagged with the _content_type
+# override and is required to keep tracking field ownership.
+
+APPLY_PATCH_CONTENT_TYPE = "application/apply-patch+yaml"
+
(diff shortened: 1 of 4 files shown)


sentry_streams_k8s/sentry_streams_k8s/operator/constants.py
```diff
@@ -0,0 +1,22 @@
...
+# override and is required to keep tracking field ownership.
+
+APPLY_PATCH_CONTENT_TYPE = "application/apply-patch+yaml"
+
+# The CR and its workloads may live in different namespaces, so we cannot use
+# ownerReferences (Kubernetes forbids cross-namespace owning) or kopf.adopt.
+# Instead every workload carries the owning CR's UID as a label with
+# name/namespace annotations so the operator can delete resources itself:
+
+OWNER_UID_LABEL = "streams.sentry.io/owner-uid"
```
fpacifici: I think this should be more specific. Let's say we add more CRDs for streams.sentry.io .
What about: pipeline.streams.sentry.io

fpacifici (approved):

sentry_streams_k8s/sentry_streams_k8s/operator/constants.py
bmcquilkin-sentry: There would never be 2 overlapping UIDs, but I can change this anyway

PR #356, opened 2026-07-29
ref(render): render pod sets or deployments

bmcquilkin-sentry:
Adds some framework to generate either Pod sets (for the operator's management) or Deployments (for PipelineStep).

### DIFF
--- a/sentry_streams_k8s/sentry_streams_k8s/consumer_builder.py
+++ b/sentry_streams_k8s/sentry_streams_k8s/consumer_builder.py
@@ -5,14 +5,47 @@ import json
 import re
 from dataclasses import dataclass, field
 from importlib.resources import files
-from typing import Any, Mapping, cast
+from typing import Any, Mapping, NotRequired, TypedDict, cast
 
 import yaml
 
+from sentry_streams_k8s.constants import CANARY_WORKLOAD_SET, PRIMARY_WORKLOAD_SET
 from sentry_streams_k8s.merge import ScalarOverwriteError, deepmerge
 from sentry_streams_k8s.validation import validate_pipeline_config
 
 LOG_LEVELS = ("DEBUG", "INFO", "WARNING", "ERROR")
+
+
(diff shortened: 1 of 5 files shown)


sentry_streams_k8s/sentry_streams_k8s/operator/constants.py
```diff
@@ -12,6 +12,10 @@
 
 APPLY_PATCH_CONTENT_TYPE = "application/apply-patch+yaml"
 
+# Identifies independently reconciled Pod sets owned by the same pipeline:
+
+WORKLOAD_SET_LABEL = "streams.sentry.io/workload-set"
```
fpacifici, line 17: This is unused

sentry_streams_k8s/sentry_streams_k8s/consumer_builder.py
```diff
@@ -5,16 +5,49 @@
...
 
 
+class PodTemplate(TypedDict):
+    spec: dict[str, Any]
+    metadata: dict[str, Any]
+
+
+@dataclass(frozen=True)
+class WorkloadSet:
+    """Group of identical replicas (primary or canary set)."""
```
fpacifici, line 26: Do you mean pods of the same pipeline with "identical replicas" ?

fpacifici (approved):

sentry_streams_k8s/sentry_streams_k8s/operator/constants.py
bmcquilkin-sentry: Artifact from splitting the CR, this is used in a future PR so I will leave it here.

sentry_streams_k8s/sentry_streams_k8s/consumer_builder.py
bmcquilkin-sentry: A workload set is just a subset of all replicas for a CR. There are 2 workload sets (primary and canary).

PR #357, opened 2026-07-29
ref(operator): daemon-based reconcile

bmcquilkin-sentry:
Uses daemon-based reconciliation and adds type stubs for the k8s objects instead of dict[str, Any].

### DIFF
--- a/sentry_streams_k8s/sentry_streams_k8s/operator/operator.py
+++ b/sentry_streams_k8s/sentry_streams_k8s/operator/operator.py
@@ -172,7 +172,7 @@ async def _reconcile_once(
                         await asyncio.to_thread(
                             _patch_pipeline_status, name, namespace, status_patch
                         )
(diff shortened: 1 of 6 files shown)


sentry_streams_k8s/sentry_streams_k8s/operator/operator.py
```diff
@@ -1,59 +1,240 @@
...
+                        logger=logger,
+                        status=status_patch,
+                        previous_conditions=previous_conditions,
+                    )
+                except kopf.PermanentError as e:
+                    if status_patch:
+                        await asyncio.to_thread(
+                            _patch_pipeline_status, name, namespace, status_patch
+                        )
+                    logger.error("%s", e)
```
fpacifici: logger.exception otherwise we loose the stacktrace.

fpacifici (approved):

PR #360, opened 2026-07-30
feat(streams): Add a fake Kafka-less source for profiling pipelines

fpacifici:
Adds a profiling mode that runs a pipeline end to end without a Kafka broker. A new `FakeSource` primitive synthesises messages of random bytes at a configurable rate and terminates after a fixed count, so the cost of the pipeline machinery itself (Python/Rust step execution) can be measured in isolation.

```python
pipeline = fake_streaming_source(
    name="fake_source",
    message_size_bytes=1024,
    messages_per_second=1000.0,
    num_messages=10000,
)
```

The idea is to instantiate the Arroyo processor with a fake consumer class instead of making the processor create its own consumer.
This can be extended into other types of sources including mock consumers for unit tests without kafka.

The fake source, right now, is pretty bare bone, it does not do anything regarding assignment, partitioning and rebalancing. This is ok as this is meant to profile the pipelines under steady load.

**Dependency note for review**

We were picking up the latest version of arroyo,  though a breaking change was introduced in https://github.com/getsentry/arroyo/pull/540, so we need to limit the version and adapt to the change in another PR.


tryangul (approved):
Somewhat of a rubber stamp, but LGTM. I think I can plugin to `fake_streaming_source`

evanh (commented):
I'm not sure I understand the use of this, since the generated messages can't be processed by any pipeline steps. Wouldn't the generated output need to be structured, so that it can be processed by the steps in the pipeline?

fpacifici:
> I'm not sure I understand the use of this, since the generated messages can't be processed by any pipeline steps. Wouldn't the generated output need to be structured, so that it can be processed by the steps in the pipeline?

Good question. Originally I wanted to plug in a system to generate valid traffic. The PR was becoming too big with the infrastructure around the fake consumer that I kept it minimal with random payload. 
This can be used to profile the wiring of the arroyo adapter (python to rust transforms) which are a problem in their own.
Then we can build on top of fake source.

evanh (approved):
> This can be used to profile the wiring of the arroyo adapter (python to rust transforms) which are a problem in their own. Then we can build on top of fake source.

If the data is random though, then every step of the pipeline will fail if it tries to process the message (which most do). So the wiring won't be tested correctly?

As you said though, we can merge this and then build on top of it.

fpacifici:
> If the data is random though, then every step of the pipeline will fail if it tries to process the message (which most do). So the wiring won't be tested correctly?

Yes you would need a pipeline suitable for this to run this test. This, at this stage, is not meant to test existing pipelines. We need to build on top of this to do so.

